PR #80, opened 2021-06-03
Make Dog Relationship type multi-select

### LINKED ISSUE #72: Add approved CEUs
- We need 1 more field, Approved CEUs (a number)
- CUE Authority should be CEU Authority

Contact Related CEU: I think we want to have it display:
 
Date Completed, Program Title, CEU Authority, Req CEUs, Approved CEUs, Status

We are going to need to be able to flag on a Trainer’s contact page if they have sufficient CEUs or don’t when it is renewal time, which means based on the trainer’s certification date, we’d want to see we have at least 12 in that 1 year period based on CUE Completed Date. We can do that w/ workflow/trigger sorts of things so I think it makes more sense there than to add anything else to the CEU object.

deb761:
# Changes
Make Dog Relationship type multi-select
Changed DogRelationShip and BackgroundCheck name fields to auto number

# Issues Closed
#72

### DIFF
--- a/force-app/main/default/layouts/Dog__c-Dog Layout.layout-meta.xml
+++ b/force-app/main/default/layouts/Dog__c-Dog Layout.layout-meta.xml
@@ -246,13 +246,7 @@
         <fields>CORE.USERS.FULL_NAME</fields>
         <relatedList>RelatedActivityList</relatedList>
     </relatedLists>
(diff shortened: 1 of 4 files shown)


atlasjen (approved):
pushed to scratch org. see i can add multiple handlers, as an example. approving.

PR #86, opened 2021-06-16
Bug/81 name fields

deb761:
# Changes

Updated recipes and sample data


atlasjen:
deployed w/o issue. still need to update lead pagelayout to include all fields, but i will provide that.

atlasjen (approved):

PR #112, opened 2021-07-24
Implement PAT object

### LINKED ISSUE #107: Implement "PAT" object
see Updated Schema spreadsheet.

deb761:
# Changes
Called PublicAccessTest__c
Ordered layout
Added related records to Contact and Dog layouts
Added fakery recipe to create one PAT per dog

# Issues Closed
#107


atlasjen (changes requested):
Things are looking good.. we need to be able to add the actual PAT as an attachment and associate it w/ each PAT record.  I assume if we add them in here, they aren't in the related "files" and that is ok.. but when we look at PAT history for a client or dog, we also need to have the physical record.

Rules/Triggers we will want
- PATs are done quarterly, so they are due by the end of the same quarter the following year. [do you want a derived field to store the date for the end of the quarter the PAT is due?]

-  1 mo before the quarter, we should have an autoemail go to the clients reminding them the quarter is coming up and ask if they want addl help for a refresher
- At the start of the quarter, task on Molly (admin assistant) to reach out to clients to discuss scheduling PATs.
- If PAT not completed 1 mo before end of quarter, some way to have it be obnoxious for the admin that this is due
- 1st 3 PATs annually. After that, every other year. I know we list video as a location. Technically that isn't allowed the 1st year, but I don't know if we want to add that logic in or not. It also isn't allowed 2 years in a row.. but again, that is business process that we don't have to capture right now
--- note, when we do the PAT we also need to ensure all vet records are up to date
--- after the PAT, we also have a client certification agreement.
I don't think these last two belong in this issue, but as they are related and on my mind, noting them. For our 1st release, I don't think we do anything special to track all of these intersecting requirements. For subsequent, i think we look at adding more business logic in, or reports in that identify where we don't have all requirements met by clients and dogs

atlasjen:
all good. decided to work w standard related files. merge once conflicts resolved

PR #119, opened 2021-07-29
Feature/email templates

SandeepSF461:
Email templates.


atlasjen (changes requested):
HI Sandeep. I am testing from w/in salesforce itself, going to Lead or Contacts, choosing New Email and then choosing to apply a template.

The templates all come up and so does the wording.
Populating names:
-the person's name populates in Lead but it does not in Contact

I have some minor wordsmithing changes I can do but that is the main thing I see at first.

One thing that is interesting. In Lead, when I click on email, it brings this up w/ the person's name included already:
image

However, in contact it does not.  I expect that might be part of the issue. thoughts?
image

SandeepSF461:
Hi Jen, So from our last call you have mentioned this email's are intended from leads so whenever you are sending these emails to leads it would automatically bring their name into the template as it is appended with "Hi {lead.firstname} ". But I have not included the contact name. 

Again correct me if I am wrong. I am happy to go back and change. Let me know jen.

atlasjen:
> Hi Jen, So from our last call you have mentioned this email's are intended from leads so whenever you are sending these emails to leads it would automatically bring their name into the template as it is appended with "Hi {lead.firstname} ". But I have not included the contact name.
> 
> Again correct me if I am wrong. I am happy to go back and change. Let me know jen.

relooked at these all.. and you are right.. these all will be leads. I wasn't sure if the teams set in motion one might be a contact..  let's see if that could also apply for contacts.

atlasjen:
> > Hi Jen, So from our last call you have mentioned this email's are intended from leads so whenever you are sending these emails to leads it would automatically bring their name into the template as it is appended with "Hi {lead.firstname} ". But I have not included the contact name.
> > Again correct me if I am wrong. I am happy to go back and change. Let me know jen.
> 
> relooked at these all.. and you are right.. these all will be leads. I wasn't sure if the teams set in motion one might be a contact.. let's see if that could also apply for contacts.

ok.. post feedback from sandeep that it needs to be Lead or Contact, we are sticking w/ Lead.

PR #123, opened 2021-08-03
file upload new Cmp

kaurarjinder:
# Critical Changes. - New component to upload file

# Changes - new picklist fields to content version

# Issues Closed


atlasjen (changes requested):
Hi Arjinder
A couple of things
1. can you fix the typo on the component name? it is presently "documetUploadCmp". vs "documentUploadCmp" in a few file names
2. How do i see this so I can test? Can you add it to the "AtlasContactLayout"  under Contact?
3. you asked about delete capability. yes, we def do want to be able to see and delete files once uploaded.

atlasjen:
hi Arjinder. Nice progress.
Still on the todo
1. fix the typo in the component name itself to "documetUploadCmp". vs "documentUploadCmp
2. delete works, but it doesn't seem to auto refresh to remove it. I got the modal saying success but had to then refresh again.. can we control that?
3. In the dialog 
- Change "Position Files" to "Position" in the Upload files dialog 
- fix the values in the user type picklist (remove "Status" from each) and go ahead and remove "Lead Facilitator" that is the same as facilitator for files.
- can you make the vertical size bigger so by default all items show so you don't have to scroll w/ the drop down?
image

4.I looked through the file types. 
- On the client one, change "veteran paperwork if applicable" to "Veteran Paperwork"
image
-On the trainer (and same on the Team Facilitator )one, Please capitalize the L in letter and change the last one to "Trainer Infraction Notice"
image

thanks!!!

PR #128, opened 2021-08-08
Feature/95 expected completion

### LINKED ISSUE #95: tasks based on Program Assignment dates
We have the optional field "ExpectedCompletion__c" let's make that required.
Can we have some default values based on type of program?

Program | Duration (months)
------------ | -------------
Volunteer | 0.5
Team Facilitator | 4
Trainer | 9
Client | 6
Teams Set in Motion Client | 6
Sweet Snoopers Trainer | 4

then, we would want to have a task go on Molly (overall program admin )
that if we are w/in 1 month of the above, a task is assigned to check on their status [for volunteer, maybe we check in after 1 week]
As we add more people monitoring folks, i could see these getting assigned to diff admin volunteers dep on the program. for now we can make it simple and be just to the main program admin

deb761:
# Changes

Created trigger to fill in expected completion date
Created custom metadata to contain program durations
Created queue group to notify of expected completion
Created scheduled apex class to create notification tasks

# Issues Closed
#95


atlasjen (changes requested):
Looks like getting null pointer trying to access progAssign.AssignedDate_c... 
image

deb761:
We were missing the duration for Volunteer.  Added a check to set duration to 0 if the duration for the program was missing.

atlasjen:
merging. issue is resolved. we should add duration for the volunteer program length. i think we said 2 weeks.

PR #154, opened 2021-11-01
Go back to direct references in log object

deb761:
This is simpler, and makes sure we have the one-to-many relationship for client, submitter, and facilitator

# Critical Changes

Removed ContactLog object and and extra LWC

# Changes

1. Updated LWC for relatedLogs and logForm.
2. Shortened method names in LogController
3. Wrote unit tests for LogController

# Issues Closed


atlasjen:
just reviewing. works better than the old one on devint! :)
There are some kinks in New and Edit related to the checkboxes for role. 

for example

1. If I click Edit, then the checkbox that was shown for Role isn't checked
2. On a given log someone would be a client a submitter or a TF, not all of the above. Would that make it easier w the logic of what to show? Right now to see all of the client options from the form for example, it seems i have to check Client and TF.
3. I can do more testing.. let me know thoughts on the multiselect implementation before i go to far. I want to make sure i understand your implementation first.

deb761:
Yes, the checkboxes are a problem.  I considered not having them at all and just showing all the fields.  Anyway, the idea is that the contact id would be assigned to the checked roles, but I'm not filling in the roles on edit since it would require loading in the record first, or getting the role list from the related logs.  Mmm.  That's an idea.

atlasjen:
merging w/ known issues on "new" form.

PR #156, opened 2021-12-05
Bug/related objects

### LINKED ISSUE #89: background check object and process- backend
**OBJECT**
We want to be able to track when a background check was requested and then the date we entered it was completed.. this lets us have a way to have a reminder if it was requested but not completed. I would assume we use the system default created date for the request date. 

1. Change "date" field to "completed date".

2. Change Result to Status and add "Pending"

3. Allow attaching file to the BackgroundCheck object. (Update fileUpload).
4. We presently download and store the background check file in sharepoint. I am thinking we would want to be able to add that as a file attachment here. The permissions would need to be restricted so only people w/ certain permissions could access the file or even see the background check results. That would impact both the Background Check object and the Contact page layout they are associated w/. Please add a "Background Check Report" field to the Background Check object.

~~(4) I am thinking we want to add a way to not have background checks always reoccur. so either a "one time" or "reoccurring" whatever you think.. but it should be in the logic below about whether to rerun. Not sure if we need this use case so maybe we hold on that until we do.~~

**INITIATING**
(1) we want to automatically initiate background checks
- when a new contact is being onboarded and picks one of the values under "Volunteer Access" that we list as requiring a background check.
-> @SandeepSF461  this would be autoemail to background check coordinator. 

-> also auto task weekly to BG check coordinator to upload result if completed date is not set AND status is not set.. this should repeat every week until cleared. again, to question above, maybe we don't need 'completed date'.  For my purposes, a record w/in a week or so is close enough.. so having it rerun based on the previous requested date is fine. International ones can take a couple of weeks but most are 1 day.

-> we want to then check their status after 1 year and rerun it if their status is active and they still have one of these roles (and it is "reoccuring"). so auto email to background check coordinator much like the 1st time around and a new task created, much like the first time around.

(2) There should be some way to manually kick off a background check. Would that be from the contact page layout? 

** CONTACT **
What do we want to show in contact? We have the fields
"Criminal conviction" and "Criminal conviction notes"
Do we drop these? we have the related data in background check. can we query it if we want to run a report? thinking maybe we drop these fields. I think the idea was they were a quick view of the latest report. Alternatively, these are autopopulated w/ "YES" if the background check fails and whatever the notes are. @deb761  thoughts?
Reality is this has not come up often and it is protected info.. i would want to be able to run a report that let me know who had criminal convictions as a summary.. or to be able to look at it person by person.. but really we don't need the detail that easy to access otherwise. **YES. We will drop. Jen to include in edits on contact.**

deb761:
# Changes

Changed role selection to radio buttons.
Update roles for edit.
Fixed problems with Apex unit tests.

# Issues Closed
#89


atlasjen:
background isn't clearing out the memory handle to previously uploaded backgound checks.
I uploaded one file and then saved 2 more backgorund check entries that show the data is persistent across objects.
I did not choose a file to upload after the 1st attempt, but I got it.

image

atlasjen:
i submitted some new issues but closing this one so we can see if the apex tests are now happy.

PR #187, opened 2022-04-10
Make log form closer to web form

### LINKED ISSUE #158: Logs fixes
1. Logs let you submit mostly empty records
Logs only require a date be set and everything else can be empty.
We should have minimum info that is required consistent w the webform.
All fields are required other than 
- handler
- life satisfaction type questions, details
2. Radio buttons don't bring up the satisfaction question options based on which radio button is selected
I only see "life satisfaction"

Below is what the webforms look like:
Team Facilitator
https://www.tfaforms.com/4819325
image

Client
https://www.tfaforms.com/4821671
image

deb761:
Made fields required and added missing fields to form.

# Issues Closed
#158

image

image

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -47,8 +47,8 @@ flows:
                 task: assign_permission_sets
                 options:
                     api_names: Atlas_Super_User
-            3:
-                task: load_dataset
+            # 3:
+            #     task: load_dataset
     config_dev:
         steps:
             1:
--- a/force-app/main/default/lwc/logFormCmp/logFormCmp.html
+++ b/force-app/main/default/lwc/logFormCmp/logFormCmp.html
@@ -6,93 +6,144 @@
         onclose={handleClose}
     >
         <div slot="body">
-            <lightning-radio-group
+            <lightning-checkbox-group
                 name="role"
                 label="Role"
                 options={options}
                 value={roles}
                 onchange={handleChange}
             >
-            </lightning-radio-group>
+            </lightning-checkbox-group>
             <lightning-record-edit-form
-                object-api-name="Log__c"
+                object-api-name={objectName}
                 class="slds-scrollable"
                 style="height: 25rem"
                 layout-type="full"
                 record-id={recordId}
             >
-                <div class="slds-grid">
+                <div class="slds-grid slds-wrap">
                     <div class="slds-col slds-size_1-of-2">
-                        <!-- Your lightning-input-field components here -->
                         <lightning-input-field
                             field-name={fields.date}
                             class="category slds-p-bottom_medium"
                         ></lightning-input-field>
+                    </div>
+                    <div
+                        if:false={isSubmitter}
+                        class="slds-col slds-size_1-of-2"
+                    >
+                        <lightning-input-field
+                            field-name={fields.submitter}
+                            class="category slds-p-bottom_medium"
+                        ></lightning-input-field>
+                    </div>
+                    <div if:false={isClient} class="slds-col slds-size_1-of-2">
+                        <lightning-input-field
+                            field-name={fields.client}
+                            class="category slds-p-bottom_medium"
+                        ></lightning-input-field>
+                    </div>
+                    <div class="slds-col slds-size_1-of-2">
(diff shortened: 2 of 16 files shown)


atlasjen:
I tried to build this to a fresh dev org and it failed, i believe in pushing data, but please confirm before i do a merge.
                    ------------------------------------------------------------
                    Running task: dx_push
                    ------------------------------------------------------------

                    Options:
                      command: force:source:push
[04/14/22 22:05:06] Beginning task: SFDXOrgTask

                    Running command: sfdx force:source:push -u test-htcwmoaec25n@example.com
[04/14/22 22:05:08] *** Pushing v50.0 metadata with SOAP API v53.0 connection ***
[04/14/22 22:06:13]
[04/14/22 22:06:14] === Component Failures [2]
                    Type   Name            Problem
                    ─────  ──────────────  ───────────────────────────────────────────────────────────      
                    Error  AtlasSuperUser  You cannot deploy to a required field: Log__c.OtherHours__c      
                    Error  Admin           You cannot deploy to a required field: Log__c.OtherHours__c      

[04/14/22 22:06:15] Return code: 1
                    stderr: ERROR running force:source:push:  Push failed.

                    Exception in task config_dev.dx_push

Error: Return code: 1
stderr: ERROR running force:source:push:  Push failed.

atlasjen (changes requested):
Please see comment re error.

atlasjen:
went to test and indeed, the deploy failed. the atlas org is not there. can't test w/o this fix.

atlasjen:
super snazzy form and logic!!
I Did some testing.
1) SessionType is still "required" even if TF is not chosen as a type and thus no submitter or client only logs work.
It does properly show if TF is selected and properly submits for the TF case. It is properly not shown for the client. Logic just needs to be changed so that it is not required unless TF is checked.

image

2) I don't think the combo of submitter, client AND TF is valid.  I think these combos are valid
- client
- submitter
- TF
- client & submitter
- TF & submitter
That said, i'm trying to think if it hurts anything to leave it there vs making the logic more sophisticated.
One thing to do to help make things clear in this case is change the label on  "life satisfaction" to "Client Life Satisfaction" make it clear it is for the client.
image

3) Trying to think of a label that makes it a bit clearer what this is. This is hours they've trained w the client since the last report.
image

4) This may be by design but when i click NEW or EDIT on a log, i briefly get a popup that has just the 3 roles shown and then it brings up the edit or create form w the full detail.

5) I don't suppose it is possible to pre-filter that only dogs related to that client are shown?  i was starting to think we would filter to only clients associated w a TF but that can vary so i don't think we do that.  Maybe pre-filtering the dogs is a 1.1 feature.

deb761:
1. I think I can take care of SessionType.
2. I can't change the label based on whether the submitter is a facilitator or not.  We'd need to create another field.
3. How about 'Hours with Client'?
4. That's JavaScript waiting for the actual record, or lightning stuff going on.
5. That filtering is more than I think we need to fuss with for now.

atlasjen:
> 1. I think I can take care of SessionType.
> 2. I can't change the label based on whether the submitter is a facilitator or not.  We'd need to create another field.
Right, so it would always be "Client Life Satisfaction"
> 3. How about 'Hours Worked with Client'?
How about "Hours worked with client since last report" ?
the idea is to capture hours if they haven't put a report in lately. one of the orgs we are certified through requires clients to have a certain number of training hours
> 4. That's JavaScript waiting for the actual record, or lightning stuff going on.
OK
> 5. That filtering is more than I think we need to fuss with for now.
YEP!

atlasjen:
@deb761 is this ready to review?

deb761:
Yes, it is ready to review.

atlasjen:
the Session Type issue is resolved.

just need these label changes that are outstanding and we are good to go
1. Change "Life Satisfaction" to "Client Life Satisfaction"
2. Change "Hours Worked with Client' to "Hours worked with client since last report" 

With that, good to merge.

deb761:
Turns out "Hours worked with client since last report" has too many characters.

PR #188, opened 2022-04-10
Add value to Programs

### LINKED ISSUE #177: Program Assignment updates
Just minor tweaks
* add "Assistance Dogs Set in Motion Client" to program assignment. program  picklist
* can we remove Program Assignment Name from Program assignment view?
image
*Related to Contact should be "Assigned Program**s**" not  "Assigned Program"

deb761:
Changed header to related list for Programs
Added bottom border to related objects.  Not perfect (radius is still there), but much better.

# Issues Closed
#177


atlasjen:
A few updates. It look like Contact Information was removed/not added to all. Each type (other than dog) should have Contact Information as a type.

Issue #66, opened 2021-04-08
Define & Update Contact Layout


atlasjen:
Jen: as part of schema review, complete mockup 
then assign to Arjinder to do the layout update

atlasjen:
This will end up mocked up to do a LWC. There are a handful of open questions running around in email threads about what need to be derived fields vs what can be queried realtime. Once resolved I will create the mockup and assign.

atlasjen:
start of notes:
**Schema**
- I think we can drop Trainer CEU Status and query the related CEU status.
- board member accepted date could be derived from Board Agreement File upload potentially
- Criminal conviction would be derived from background check vs set explicitly
- Drop criminal conviction notes


**Layout**
- In related, remove: opportunities, Cases, Campaign history

- In related, change" Program Assigned" to "Program**s** Assigned"

- In filter view, change "doner" to  "donor"

- Swap so details tab is first and related tab is second

- we will want to group the fields that only pertain to a given constituent. so Trainer, client, etc. 

- it would be nice to see a display in the header of status on each position the contact may have

- drop "Add to Campaign"

**New Contact Form**
In general contacts will be made 
- converting from leads
- by FA if we get applications
- data migration/importing data
The only time we would manually create a new contact is if we know they are going to move forward and we don't want to bother making a lead and converting.. or do we hide the ability to make a new contact at all??

atlasjen:
keeping open as we narrow down on final schema and page layout for contact.

atlasjen:
done enough for this release.

PR #191, opened 2022-04-16
Bug/layoutupdate

### LINKED ISSUE #87: update lead page layout to match lead input form
Can you update the Lead page layout to include all data in the lead NEW form and tweak cosmetically? I like how you've been doing layouts, so use your judgement.
let me know if you need other guidance.

atlasjen:
# Changes
Updated layouts so 
- preferred tab is always "Details"
- removed things like chatter and news
- updated Contact Layout to have cleaner section data
# Issues In Progress
#66 
# Issues Closed
#87

### DIFF
--- a/force-app/main/default/flexipages/BackgroundChecks.flexipage-meta.xml
+++ b/force-app/main/default/flexipages/BackgroundChecks.flexipage-meta.xml
@@ -1,10 +1,15 @@
-<?xml version="1.0" encoding="UTF-8" ?>
+<?xml version="1.0" encoding="UTF-8"?>
 <FlexiPage xmlns="http://soap.sforce.com/2006/04/metadata">
     <flexiPageRegions>
         <name>header</name>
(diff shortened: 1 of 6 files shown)


deb761 (commented):
I notice related Background Checks are in the Contact layout in both the layout and the record page (as the relatedBackgroundChecksCmp).

The reason for the relatedBackgroundChecksCmp was to override the New & Edit buttons.

As I recall, I added overrides for the new and edit background check buttons, so we might be able to dispense with the relatedBackgroundChecksCmp.  Otherwise, we should remove the related list from the contact layout.

atlasjen:
i expect i accidentally added the wrong component. interestingly it seems like all of the related items i see on our present contacts are not listed in the Atlas Contact Layout.

My intent wasn't to change anything in related other than removing default components we weren't using. Can you fix?

deb761 (approved):

PR #193, opened 2022-04-16
157 background check lets you create empty records

### LINKED ISSUE #157: Background check let's you create empty records
If you create a new background check, there are no required fields.
It will create a record w the default status and then also adds an invalid file handle.
image

If i click on the file, I get this error
image

deb761:
# Changes
Add requirement to have an associated file before creating or updating a background check.
Disable file view when there is no file.

# Issues Closed
#157


atlasjen:
Can you also make date required? Not sure if you can show a "*" next to Date and "Add File" I know you have the logic to make sure the file is there. With this change, ready to commit.

atlasjen (changes requested):
see note on minor change to also make date required and see if you can put * by that and file.

deb761:
I can't make the file insert required because its in a sfdc lightning component.

atlasjen:
> I can't make the file insert required because its in a sfdc lightning component.

A bit confused what you mean.. do you mean the date?

atlasjen:
yay! happy.

PR #196, opened 2022-04-21
185 add v4s volunteers for salesforce

### LINKED ISSUE #185: Add V4S - volunteers for salesforce
In researching, i think the free volunteers4salesforce app will get us a lot. It was actually installed in our old instance but i didn't understand what it did.
More data is here, including their data dictionary.
https://powerofus.force.com/s/article/V4S-Data-Dictionary

For this 1st release my question is, can we add it and in this release all we are doing is using their custom volunteer related fields instead of our own.
We would replace these Contact fields we created:
Volunteer Background  -> their field Volunteer Skills
Volunteer Focus  --> their object Volunteer Job
Volunteer Interview Notes  -> their field Volunteer Notes
Volunteer Status --> their field volunteer status

The only thing it doesn't include is background check info (volunteer access) so we can just keep that the way we have it... which is great as that is the one thing we did that is custom on volunteers!

It does add fields to Lead, so sandeep would need to map to those for the contact us form, but that's fine. they will automatically convert when converted to a contact.

Then in subsequent releases we can figure out a bunch of other website integrations/using it/volunteer tracking etc.
I would just rather do the import to the fields once and this gives us a way to solve a lot of the volunteer mgmt issues we are trying to figure out.

Package deployment implications?

deb761:
# Changes
Updated fields and sample data

# Issues Closed
#185


atlasjen:
looks good!

PR #197, opened 2022-04-22
Feature/190 dog fields

### LINKED ISSUE #190: Dog fields
For display purposes lets add

*Age (Years)- this would be a decimal like 1.5
I'd like to be able to show this in the list of all dogs vs months
image

*Health Care Record Status Summary- 
   we used to have this, essentially a rollup of what was in the health care record status field. If everything in it was good, then the summary is good. if any one thing is in "due soon" stage but rest was good, then status is "due soon". if any one thing in "late" stage, status is "late".  I'd like to be able to add this to the top level detail for Dog next to where it has the overall status so it has Status and Health status summary both.

image

deb761:
# Changes

Restore  ExpiryDateStatus to Dog Highlight Panel.

# Issues Closed
#190


atlasjen (changes requested):
see comments on #190 that answer remaining questions.

atlasjen:
love the graphics! perfect!

PR #206, opened 2022-04-26
184 create receiveddate field for each file upload type

### LINKED ISSUE #184: Create ReceivedDate field for each file upload type
maybe do this at the same time as #181 
for every file type, we need to populate a ReceivedDate field so we can use it for reporting.

I suggest a format of <category><document type>ReceivedDate
So for example, ClientPreApplicationReceivedDate, VolunteerEmergencyContactReceivedDate

these should be populated w a trigger anytime one of these files is uploaded.

If the file isn't given both a category and type then we don't set a date.

deb761:
# Critical Changes
Created fields for Received date for File Category/Type

# Changes

# Issues Closed
#184


deb761:
I just realize I'll need to update the sample data for this one.

atlasjen:
Please add these back in "ContactForm" on Volunteer, Board, Trainer, Client, Facilitator, Staff  and their associated Received Dates.
Please remove TSiMPreApplicationReceived

atlasjen:
Tested a few cases
- added a date manually. uploaded a file, it correctly updated the date.
- uploaded a file. deleted a file. It did not clear the date. thinking if a file is deleted, it should drop to the most recent of the same type and if none other of that type, should be deleted
- do we change the fields in the page layout to be read only? Thinking we do. 
- can we add these fields to the dog layout? i believe it is a custom component so not sure how to edit it myself

atlasjen:
For this release
- known issue if you edit/delete a file, the date will be incorrect
- known that people can manually edit dates
Next release
- add triggers so it will work w/ FA as well as file changes
Only thing to do is add these received date fields to dog Jen to work w arjinder.

deb761:
Another thing to note is that when you update a Program status the contact status doesn't update: you need to refresh the page.
I can put together a trigger to update that for 1.1.

PR #210, opened 2022-04-28
Contact status rollup

### LINKED ISSUE #195: contact status updates
1) I am trying to think of how to capture of some our additional statuses w/o having them infinitely expand.
I know you added File types for TSiM and ADSiM vs having them under Client.
We also are opening our training platform to let people take individual programs vs do our certification program.
I am currently thinking "Other Programs" to be the catch all until we see what really expands, though TSiM is def one that is expanding and people taking indiv trainer programs will too. 

2) We need a graphical summary of all statuses that we had on contact so we can have it in the title bar for a given contact or in the contact details. We had this and it was removed in our update to this datamodel.
The idea of this is much like what we have for Dog Health Status where we have different states and it is derived.
Here is one suggestion (pending what is decided on 1 for statuses)
image

and "other programs" would cover ADSiM, TSiM and academy programs.. 
let me know your thoughts and then i can write up the criteria for each stage for each type.

deb761:
# Changes
Updated the icons we are using for status.
Added status of "other" programs.
Updated the permission set and profile.

User will need to refresh page to see updated status rollups if updating a status on the page.

# Issues Closed
closes #195


atlasjen:
It looks like these are the 2 classes that have the logic for what to show:

app/main/default/lwc/contactStatusIndicatorCmp/contactStatusIndicatorCmp.js
force-app/main/default/lwc/contactStatusCmp/contactStatusCmp.js

are we able to use a mix of dates of files, dates in the program assignments and the status on the contact? Let me know and i can give you updated info for the criteria, esp for Needs Action and Needs Action Soon.

Can you also remember to add the Last* fields to Dog?

deb761:
I certainly need the logic for Needs Action & …Soon. 

Yes, I can add the fields to the dog layout.

atlasjen:
OK let's try this logic out. Let me know what you think. I'll put the icon/display status 1st and the criteria next for each position.

**Volunteer**
just match the status in GW_Volunteers__Volunteer_Status__c except we do need some way to say truly inactive.. wondering if we need another status of Concluded vs Inactive? Some people start and stop. I think we do that and then we use the None icon for Concluded... in other words, we would want to not list them as a volunteer up here.

**Client**

- Initial Inquiry- Client Status = Initial Inquiry
- Onboarding - Client status = Onboarding
- In Training - Client Status = In Training OR In ADSiM
- Active - Client Status = Certified  
- Action Needed-Client Status = Certified AND  ClientCertAgreementReceived__cc >=1 year old 
- Action Needed Soon -TF Status = Certified AND ClientCertAgreementReceived__c >= 9 mo old
- On Hold- Client status =On Hold. 
- Suspended- Client status =Decertified
- None- Change Inactive status to concluded to be consistent. Inactive doesn't make as much sense for a client since we have On Hold.

**Trainer**
- Initial Inquiry- Trainer Status = Initial Inquiry
- Onboarding - Trainer status = Onboarding
- In Training - Trainer Status = In Training 
- Active - Trainer Status = Certified  OR ACTIVE
- Action Needed-Trainer Status = Active AND  TrainerCertAgreementReceived__c >=1 year old 
- Action Needed Soon -Trainer Status = Certified AND TrainerCertAgreementReceived__c >= 9 mo old
- Suspended- Trainer status =Suspended
- Inactvite - Trainer status =Iinactive
- None- Add "concluded" as status to be consistent.


**Staff**
N/A for now unless we want to include background check related stuff here. I say we leave that out of 1.0

**Donor**
N/A for now

**Board**
So the big thing we want to track is if they are going to need to renew soon or need to update paperwork.
For 1.0, let's keep it easy and go with matching BoardMemberStatus__c
- None/Concluded- Again, like volunteer, I think we need to add a status of Concluded. in other words, we would want to not list them as a board up here.

**Team Facilitator**
Just use Team Facilitator and ignore TF Lead for 1.0

- Initial Inquiry- TF Status = Initial Inquiry
- Onboarding - TF status = Onboarding
- In Training - TF Status = In Training
- Active - TF Status = Certified 
- Action Needed-TF Status = Certified AND  FacilitatorCertAgreementReceived__c >=1 year old 
- Action Needed Soon -TF Status = Certified AND FacilitatorCertAgreementReceived__c >= 9 mo old
- On Hold- TF status =Inactive. 
- Suspended- TF status =Suspended
- None/Concluded- Again, like volunteer, I think we need to add a status of Concluded. in other words, we would want to not list them as a TF up here.

**Other**
This is purely based on Program Assignment
In Training- if they have any of the following programs:
- TSiM
- Sweet Snoopers
- ADSiM
[ we are about to add more]

In Training = any of above programs AND any of their status = Inprogress
Decertifed/Suspended = any of above programs AND all of their status = Completed, Discontinued, Removed
Action Needed -any of the above programs AND status = Inprogress AND any have Expected Completion Date < TODAY and do not have Completion Date set 
Action Needed soon- any of the above programs AND status = Inprogress AND any have Expected Completion Date within 30 days and do not have Completion Date set

deb761:
I don't think we need the Concluded status if we only show the status for "positions" assigned to a contact.  So when the Board Member stops being a board member, you remove the position from the contact, and the status goes away.

atlasjen:
i'm thinking for reporting vs just what we show here that we want a way to know they are truly done vs inactive and might come back.
But yes, i guess we could just remove the position from the contact.
and then if we want a historical report of what roles someone has played.. how would we do that?
My inclination would be to add Concluded. We don't have to actually show it in the Status display, but i'd want to know historically.

atlasjen:
It may be in sample data, but it looks like we are using "Prospective" as a volunteer status but that is inactive in the picklist.
Look at Mckay Caitlyn.

Also looking at this same contact, I see status for Client, Trainer, Board but it isn't displayed in the component.

I also think i very poorly explained my ask on concluded vs None.
If a status is none, don't display anything in the component. Just like you were doing.
If a status is concluded, then display the graphic we agreed on for none. use it for that instead.

Let me know when you've had a chance to fix the above and i will test out the rest of the logic.

atlasjen (changes requested):
It may be in sample data, but it looks like we are using "Prospective" as a volunteer status but that is inactive in the picklist.
Look at Mckay Caitlyn.

Also looking at this same contact, I see status for Client, Trainer, Board but it isn't displayed in the component.

I also think i very poorly explained my ask on concluded vs None.
If a status is none, don't display anything in the component. Just like you were doing.
If a status is concluded, then display the graphic we agreed on for none. use it for that instead.

Let me know when you've had a chance to fix the above and i will test out the rest of the logic.

deb761:
For historical positions, we would have the _Status fields to look at, and also the Last.... fields.

atlasjen:
Add "Puppy Raiser" and Puppy Raiser status (use same exact values as TF status)

PR #212, opened 2022-04-29
Deactivate relationship types

deb761:
Fix contact layout

# Issues Closed
208


atlasjen (approved):
note.. when i first tried opening a couple of the contacts i got a message about there was an error and i might need to reload the page. I kicked off a fresh build as this was a dev org i'd done a few pushes to. I couldn't reproduce the issue, so likely all good. i think we will end up dropping a whole bunch more of these relationships but this is fine for now. let's see how it plays out as we add data.

Issue #259, opened 2022-07-02
Account layout/new updates for Email


atlasjen:
* New Organization Account should include "email" as an option to set.

* Account page layout 
  * should include email
  * swap so Details is the default tab and on the left vs related

deb761:
Closing with #261

Issue #249, opened 2022-06-18
Contact fields


atlasjen:
We need to be able to easily report on when a trainer or team facilitator needs to renew.
I was thinking we'd use the dates on the agreements, but we likely should have CertificationValidUntil dates 

TrainerCertificationValidUntil
TFCertificationValidUntil
ClientCertificationValidUntil  = PAT Valid Until
BoardTermValidUntil

We already have this covered for clients under the Team (PAT Valid Until)

atlasjen:
Update..
let's make some of these derived

TrainerCertificationValidUntil = date of latest certification agreement + 12 mo
ClientCertificationValidUntil = date of latest certification agreement + 12 mo

TF and Board will still be manually set.

deb761:
There's a problem with ClientCertificationValidUntil.  I can't create a formula field that derives that from Team, since it isn't a direct lookup.  I'd have to make it a regular field and set it off the Team field.

atlasjen:
Gotcha. Let’s hold on that this week . Im out of town w/o a computer. I know we have it on team, just thinking being able to access/query via client might be handy.. we will hit rare cases where silent is part of more than one team.. which is why we did team.

Once we have reporting dashboards I think we can more easily find status

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

atlasjen:
OK.. I see you added the other valid untils until with PR #246 . 
Let's go ahead and add the ClientValidUntil as you describe, but having it be triggered by what is set on Team.
then this issue can be closed.

PR #261, opened 2022-07-03
Feature/254 demographic data

### LINKED ISSUE #254: We need to add demographic data for contacts
We need to add demographic data for contacts.
We will definitely be wanting to protect access to it.
I am thinking we update the Contact Form and people can add the data there if they choose to.
**@debbie: I am not seeing any built in demographic fields so wanted to check w/ you on this and see if you see any.**

I am waiting to hear back from the people in the board who are working on this about exactly which categories we will include, but most likely:

Age
Gender
Education
Marital status
Race
Ethnicity
Religion
Language
Veteran Status
Disability Status

We are going to have people complete updated Contact Forms w/ this info and then we can use dataloader to get those in.

deb761:
# Critical Changes

Added demographic data
Added ClientValidUntil
Added Trigger on Team__c.PatValidUntil

# Changes

# Issues Closed
#254 #249 #259


atlasjen (changes requested):
* Can we rename to ClientCertificationValidUntil_c  and display label "Client Certification Valid Until"
* Is TrainerCertificationValidUntil derived? I noticed it is read-only in the page layout and i can't change that.
* Can you change the display label on Specified Pronoun to be "Specified Pronoun"  vs "Pronoun (Entered)" that keeps it consistent w the rest
***I added this into my commit but I removed Notes from the related list for Atlas Contact Layout so i could try to add it to the quick access bar on the right, but now i can't find out how to access it at all.**


All else looks good.

deb761:
I changed the field name and restored Notes to the bottom of the Atlas Contact Layout. 
TrainerCertificationValidUntil__c is derived: it is 365 days after TrainerCertAgreementReceived__c.
Set the page layouts for Account to be the Organization and Household layouts.  The Organization layout includes Email.

atlasjen:
* Organization Page Layout does include email, but it's not assigned "Account Layout" is the one assigned and also what impact the New form. It seems to me we should update the Account Layout since some accounts are household (most) and some are organization (few).. using the main account layout seems ok to me.

* Let's put this formula for FacilitatorCertificationValidUntil
FacilitatorCertAgreementReceived__c + 365
and then we can make that field locked in the page layout.

*Confirm that Notes are back in related, Confirm Client cert valid rename.

deb761:
When you create a Contact, a household account is automatically created.  When you create an Account for a Vet, it is an Organizational account.  And it doesn't look like you can create an account that isn't an Organization or a Household.

atlasjen:
merging in. Will add a separate issue for the Team Facilitator Valid until date.

PR #265, opened 2022-07-05
Update Account layout to make Details default

deb761:
# Critical Changes

# Changes

Make Details default on Account layout

# Issues Closed


atlasjen:
i'm a bit confused about what to review here. this also doesn't have the other changes you just made to the list views for account to have Household vs Organization as filter options in the list. is that just because this was pulled from an earlier version?

I deployed a new dev scratch org and tried to push this, and I get the following error:
image

So I did a fresh deploy w this as a new scratch org.

I still don't see "email" as an option if i pick new organizational account. 

Help me know what to look for.

PR #267, opened 2022-07-08
258 type does not seem to get populated on log

### LINKED ISSUE #258: Type does not seem to get populated on Log
In testing the list views, i created a couple of logs. I chose Client for one and Team Facilitator for one, but in both cases "Type" is not populated.
image

deb761:
# Critical Changes

# Changes
Added setting Record Type to log form.
Changed list views to show RecordType rather than Type.

# Issues Closed
#258


atlasjen (changes requested):
1. If I select "new" on the Log related item on Contact, nothing happens. it just silently does nothing. I assume there is an error being caught in the background.
2. In the "new" form on the Logs when accessed from the top menu bar, I can create logs. It properly lets me choose client or TF. In that form however:
* Submitter should be optional
* If it is a TF log, Facilitator should be required
* Stress seems to default to "None at all" vs "---None---"
* Can we move up where the TF box is?
image
3. Details on Log doesn't actually include the type
image
4. We should include the Client and Team Facilitator names in the column display in the Log related list.
5. 
image


-

atlasjen:
The new button is now happy. Do you have time to do the other listed tweaks?

deb761:
No

PR #287, opened 2022-07-18
Change text field lengths

### LINKED ISSUE #283: field length for some of the specify/other fields is short
I was trying to test importing leads from form assembly and realized we have pretty short character limits on some of the specify/other fields.
I'm really hoping we can still change these..... not sure what our datamodel change limitations are at this point. Sorry i just caught this.

**These max at 20 characters:**
**Lead**
Lead. atlas1__pronounOther__c  is text(20)
Lead.atlas1__LeadSourceOther__c  is text(20)

interestingly the field it maps to in contact is 120
**Contact**
Contact.atlas1__LeadSourceOther__c is text(120)
but pronoun is still 20
Contact.atlas1__pronounOther__c  is text(20)

I also see that the Specified things from demographics may be a bit short- particularly the accommodation one.
image

It seems salesforce tends to use Text(128) as its default length for single text fields like title, etc. Let's do that on all of these to be consistent.

It would be wise to go through our custom objects and see where we have "other". I think i caught the remaining ones below.

**CEU**
image

**Dog**
I like that microchip is unique case insensitive, but 30 seems short
let's just make everything 128 to be consistent, same w/ the dog's name
image

**Team**
Name maxes at 80. we should likely be consistent and make it 128, that said i don't care that much on that one.

deb761:
# Critical Changes

# Changes
Change field length for Specified... fields to 128 characters.

# Issues Closed
Fixes #283

### DIFF
--- a/force-app/main/default/globalValueSets/Language.globalValueSet-meta.xml
+++ b/force-app/main/default/globalValueSets/Language.globalValueSet-meta.xml
@@ -46,6 +46,11 @@
         <label>Korean</label>
     </customValue>
     <customValue>
+        <fullName>Japanese</fullName>
+        <default>false</default>
+        <label>Japanese</label>
+    </customValue>
+    <customValue>
         <fullName>Russian</fullName>
         <default>false</default>
         <label>Russian</label>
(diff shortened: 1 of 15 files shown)


atlasjen (changes requested):
All look good other than the microchip one which is still 30.
Make that change and we are set.
thanks!

deb761:
I thought you wanted that one to be limited.

atlasjen:
i wanted it to still be restricted to be unique but to have it be longer.

deb761:
So, what length do you want?  You implied 128 characters in the issue.  40 characters? 60?

atlasjen:
60 works

atlasjen (approved):

PR #292, opened 2022-07-21
Make GenderIdentity a multi-select.

### LINKED ISSUE #291: fixes before final release
* Contact. 
1. Change Gender Identity to multiselect
All happy w. the demographic stuff otherwise.

* Relationship
1. Ability to view All was blown away again. Please add whatever you did in prod to the always do in staging list

* Lead

1. LeadSource Other is missing from the Lead Page layout. It does properly map to Contact. Please add that to the Lead Layout.

2. FYI, The custom lead mapping fields got blown away, so i added them to the do everytime in staging list.

3. Interestingly, Salesforce does not allow Lead.Description to be mapped to anything other than Contact.Description.
https://help.salesforce.com/s/articleView?id=000325229&type=1 
This merits a discussion. I'm trying to convince myself this waits until 1.2 but it does blow away data so trying to think of options to avoid doing much with this until 1.2. is the trigger hack an option?

Options for us that I can think of for item 3 above. maybe you have a more creative idea.
1. say ok and just hide Lead Comment from the Contact Page layout [or in 1.1 ignore it] and let all go to description<- this is an issue because emergency contact information notes also map here and would blow lead comments away.. 
2. add custom Lead.Lead Comment field that we can then map to Contact.Lead Comment.. that means changes in several places [ 1.2]
3. add some trigger hack to copy what was in Lead.Description to Contact.Lead Comment at time of conversion<- hack! and remove this in 1.2  and add in the Lead.Comment field and logic at that time? <- Seems like maybe best compromise for 1.1?
4. a bit of a hack.. but i can use form assembly to put the emergency comment to the contact. lead comment field. so this would let us keep the data separate. maybe we rename contact.lead comment to contact.emergency notes? [1.1??]

deb761:
# Critical Changes

# Changes

Make Gender Identity multi-select.

# Issues Closed
#291

### DIFF
--- a/force-app/main/default/objects/Contact/fields/GenderIdentity__c.field-meta.xml
+++ b/force-app/main/default/objects/Contact/fields/GenderIdentity__c.field-meta.xml
@@ -5,7 +5,7 @@
     <label>Gender Identity</label>
     <required>false</required>
     <trackFeedHistory>false</trackFeedHistory>
-    <type>Picklist</type>
+    <type>MultiselectPicklist</type>
     <valueSet>
         <restricted>true</restricted>
         <valueSetDefinition>
@@ -52,4 +52,5 @@
             </value>
         </valueSetDefinition>
     </valueSet>
+    <visibleLines>4</visibleLines>
 </CustomField>


atlasjen:
need to run in office and will build a new pull.. it's not letting me update this in the scratch org.

PR #303, opened 2022-08-09
Set up triggers and add tests

deb761:
For update to ContentDocumentLink and ContentVersion.
The date is only updated when it is newer.  Not 100% sure this is the behavior we want.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/force-app/main/default/triggers/ContentDocumentLinkTrigger.trigger
+++ b/force-app/main/default/triggers/ContentDocumentLinkTrigger.trigger
@@ -1,0 +1,12 @@
+trigger ContentDocumentLinkTrigger on ContentDocumentLink (after insert) {
+   for (ContentDocumentLink cdl : Trigger.new) {
+        //Get Related document
+        ContentVersion cv = [
+            SELECT Id, Category__c, Type__c, Date__c, ContentDocumentId
+            FROM ContentVersion
+            WHERE ContentDocumentId = :cdl.ContentDocumentId
+            LIMIT 1
+        ];
+        FileService.updateDate(cv, cv.Date__c);
+    }
+}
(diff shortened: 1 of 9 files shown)


atlasjen:
checking this out now. 

Note, to fully test w/ form assembly i will need it in devint since i can't connect from form assembly to a scratch org.

I think since the date is "latest" it's ok if it only updates if it is newer.

One thing i noticed in earlier version is if you delete a file, the date remains set. Ideally it would go to the next oldest one or clear if there is no file there. Not sure if you implemented anything related to deleting files in this update.

atlasjen:
dates seem to set just fine and indeed they do update if you go to the newest one.
question for whether it can be in this fix
- fixing the date that shows in the LWC. It is showing the selected date in the Last Modified date fields, but not in the LWC
- changing it so if you delete a file we recalculate the lastModified date field.
let me know. once we merge i can test against form assembly in devint.

ps, molly told me some dates weren't setting properly, but in looking at her notes, she gave me no info on which ones.. so until we get more useful info we can ignore that and maybe in your testing/test cases you found the issues anyway.

atlasjen (changes requested):
see previous comment and let me know if doable here

deb761:
I should be able to refresh the contact/dog details after a file upload.  Let's put the delete in a different issue, it's a fairly big task.  I'm not sure what you mean by not showing in the LWC.

atlasjen:
OK moved the delete to a new issue.

Sorry, that should have been fixing the date so it shows properly (utc conversion/etc) in the LWC view. If I set it to 8/10 for example, in the LWC file upload component i see 8/9. I do now see 8/10 properly in the Last* dates

atlasjen:
Committing as the date issue is not a blocker and getting this in is. We can deal w/ the date issue later.

PR #310, opened 2022-08-15
Show local date in related files

### LINKED ISSUE #296: File upload saves on previous date
I noticed as I was uploading files for Dog (Charlotte) that the dates would be assigned to the previous day vs the day I entered.
I need to do a bit more testing to see if it is always the previous day or only on a weekend.

Accuracy is good enough for our purposes, but recording to examine later.

deb761:
# Critical Changes

# Changes

Changed date in related files to date-local to match entered date.
Removed update to last* files because that is in the trigger now.

# Issues Closed
#296

### DIFF
--- a/force-app/main/default/classes/FileController.cls
+++ b/force-app/main/default/classes/FileController.cls
@@ -21,7 +21,6 @@ public inherited sharing class FileController {
                 cv.Date__c = docDate;
             }
             update cvList;
-            FileService.updateDates(cvList, docDate);
         } catch (DmlException ex) {
             throw new AuraHandledException(ex.getMessage());
         }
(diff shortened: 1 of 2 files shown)


atlasjen (changes requested):
The dates look good, however, i am seeing the Last* date not set in some cases.
I test Client/LOMI and the Last* date is not displaying.
Client/TSIM application late date is mapped just fine.

deb761:
I’ll check that. Maybe the trigger is missing that one.

atlasjen:
thanks! do the unit tests have each date included?

deb761:
They are supposed to. I may have missed one.

deb761:
They do have Client/LOMI.  Not sure why it wasn't showing in the browser.

atlasjen:
ok i just tested it again and the date came through. merging.

PR #312, opened 2022-08-15
Change TSiM to Standalone

### LINKED ISSUE #305: Changes to standardize standalone programs vs TSiM
Going with what you suggested in the first place and using reports from Program Assignments to figure this stuff out. Note, this can wait until 1.2 assuming we get that out in the next few months.
We can do some data munging then. it doesn't impact that many records.

Let's
- drop TSiM dates from Contact (last TSiM Application, last TSiM program Agreement)
- drop "In TSiM" status from Contact.ClientStatus picklist
- in the file upload LWC
  - Change Category "Teams Set in Motion" to "Standalone Programs"/ disable TSiM and add "Standalone Programs" if needed
  - they still get the same two file types of application and agreement and contact form. if they use application or agreement, no Last* dates to track. Same logic applies for contact form as for anyone else, we still track that date.

deb761:
# Critical Changes

# Changes
Change TSiM to Standalone Programs in File type list
Remove Last TSiM file dates from Contact layout

# Issues Closed
#305

### DIFF
--- a/force-app/main/default/classes/FileService.cls
+++ b/force-app/main/default/classes/FileService.cls
@@ -67,10 +67,12 @@ public with sharing class FileService{
         List<Id> ids = new List<Id>();
         for (ContentVersion cv : cvs){
             string category = (cv.Type__c == 'ContactForm') ? '' : cv.Category__c;
+            if (category == 'Standalone') continue;
             string field = string.format('{0}{1}Received__c', new List<Object>{ category, cv.Type__c.replace(' ', '') });
             fields.add(field);
             ids.add(cv.Id);
         }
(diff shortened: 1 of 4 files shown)


atlasjen:
Question:
What happens for files previously uploaded as TSiM. Do they change to Standalone programs? Or do we need to change that?
Otherwise, it works as expected and we can merge, but curious about that.

deb761:
Not 100% sure, but I think they would remain as TSiM. I could write an install script to update.

atlasjen:
if we can do the data munge that would be good vs going and manually updating the files.

atlasjen:
merging. we just need to munge the data once live.

PR #313, opened 2022-08-15
Remove notes from related files

### LINKED ISSUE #302: Supress showing Notes in Files
I am finding in Prod, anytime i add a note, it also appears in the files.
Apparently this may be by design as Enhanced Notes are saved as files..

Main todo here is see if you can hide Notes from the custom Files display LWC

-----Ignore below, it happened once---
One thing that is very odd is in some cases they create duplicate files. Look at related files for Katherine Herko.

https://atlasassistancedogs.lightning.force.com/lightning/r/Contact/0034x000019u45KAAQ/view
There is just one note but it is here 3x.
image

Can you:
1. See if there is a configuration to not have notes display in files
2. see if there is some configuration that might explain why it is there 3x?  Note, this was a note added when she was a lead and then it was converted. So that may be why the duplicate.
thanks!

deb761:
# Critical Changes

# Changes

Filtered Related Files to remove notes

# Issues Closed
fixes #302

### DIFF
--- a/force-app/main/default/classes/FileController.cls
+++ b/force-app/main/default/classes/FileController.cls
@@ -123,7 +123,7 @@ public inherited sharing class FileController {
                     Type__c,
                     Date__c
                 FROM ContentVersion
-                WHERE ContentDocumentId IN :lstConDocs
+                WHERE ContentDocumentId IN :lstConDocs AND FileType != 'SNOTE'
                 ORDER BY Date__c DESC
             ];
         } else {


atlasjen:
works great with new notes. Will need to see in staging where we have notes already to see what it does to current notes. Merging.

deb761:
Oh, we will need to munge:

Hey, we encountered this problem while adding a new field into our package.
This is Salesforce official comment:
"Yes, as mentioned by you it is a limitation. Any new packaged picklist values will need to be either manually added after the upgrade or created via a postinstall script with the Metadata API."
https://partners.salesforce.com/partnerCaseDetails?Id=5000M00000s2tgvQAA

atlasjen:
not sure about why this comment is on this issue since it is about notes and related files... ??

deb761:
Ooops!  Copied it into the wrong issue.

Issue #315, opened 2022-08-16
Puppy Raiser updates


atlasjen:
Add a file type of "Contract"
and "Last Contract" date that is populated and in the page layout

deb761:
The last ask is a bit tougher, and is related to how the lwc datatable is setup: the value of the picklist is shown rather than the label.  I'm investigating, but if we're trying to get 1.3 out this month, it's pushing it.  I won't be able to work this week (1st week of September) as I have family in town.

atlasjen:
Can we change the value and the label to both be Puppy Raiser?

deb761:
Too many characters for the last received field.

PR #326, opened 2022-09-09
315 puppy raiser updates

### LINKED ISSUE #324: Files LWC updates
1. Make it so we can sort by the various columns by clicking on them. Filtering would be an even better nice to have.
Sorting by category and date will be the most needed, but i'd love to be able to do something like
- filter by category and then sort by date or
- filter by date and then sort by category

The sorting when clicking on a column is standard salesforce behavior, so we just need to add to the LWC

2. Limit how many are displayed to say 10, with an option to view all. 
See Jennifer Kolar as a contact for an example w/ a ton of files.
https://atlasassistancedogs.lightning.force.com/lightning/r/Contact/0034x00001COhAUAA1/view

deb761:
# Critical Changes

# Changes
Updated related files to show the category and type labels, rather than the picklist value.
Modified return from FileController to include total number of related files and limit the number of files returned.
Modified relatedObjects component to show the total number of objects and provide an area to select View All.
Added sorting to relatedFiles.
Created component to show all files.

TODO: add filtering to all related files.

# Issues Closed
#324 #315


atlasjen (changes requested):
Looking good!
 I think you got most stuff from the issues.

A few comments on what i see missing/left
- Please add "active" as a status for the Puppy Raiser picklist
- View all doesn't automatically refresh. is that possible? I clicked on view all when i had 1 file, came back, added more files, then clicked view all and only saw 1 file.

atlasjen:
One more comment.
Since you are using a custom component for View All, can we add the ability to filter not just sort? In particular by Category?

deb761:
I can't refresh view all for items added in the record page.  You'd want to add the items in view all.

atlasjen (changes requested):
"Active" now there for status.  :)

However, I just noticed that the Last Contract is not present. Instead Puppy Raiser Cert Agreement is there twice.  Neither one is filled out when the contract is uploaded.

image

deb761:
Copy-paste error with the label for the new field.

atlasjen (changes requested):
This has the proper label, but the value is still not getting set. I refreshed several times.

image

image

deb761:
Should be fixed now.  Once we merge this in, I can use the code generator to fix this.

atlasjen:
Merging. There is still an issue w/ Last Dates displaying, but it is separate from this . closing and will make a new issue.

PR #327, opened 2022-09-09
Fix typo so delete works

### LINKED ISSUE #325: Unable to Delete Log from w/ in Related list in contact
I tried this out in prod.
I manually added a client/submitter log to Jennifer Kolar as a test.
I went to try to delete it using the Delete option in the list and it silently does nothing.
image

Note, if i go to the list of ALL logs as an object, not as a related list, then i can delete them.

deb761:
# Critical Changes

# Changes

Fix delete within RelatedLogs

# Issues Closed
#325


atlasjen (changes requested):
Is there any way to show it is in progress to delete? right now i am not seeing any spinny/notice of anything happening and i need to refresh to see the entry is gone.

deb761:
When I test with a scratch org, I see the delete happening right away, and the page is refreshed after the delete.

PR #328, opened 2022-09-09
260 dog changes

### LINKED ISSUE #260: Dog changes
**New Dog:**
Remove Vet Clinic from New Dog form.
Add Birthdate to New Dog form.

**Datamodel:**
Deprecate "Activation Date"
Deprecate "Expired"

**Page Layout**
Fix tool tip on birthdate to be MM/DD/YYYY or mm/dd/yyyy something consistent. might as well do the same thing you do for #314 

image
-------------------

History:
If I use the "new" dog form, when i search by vet clinic, it does not give me the option to create a new account. Which is weird, since all of the other new forms i see by default let you add new contacts if you can't find one. maybe this is because it is an account vs contact?

image

Then, let's say i pick an existing account as the vet clinic. Now, I come to look at Related humans and no human is listed:
image

Semi related, What if a vet clinic has 2 vets and one client uses one and 1 uses a diff one.. so we have 1 organization account but 2 contacts. How do we specify which vet is the related human? I know we wanted to support the case where they had a clinic vs a specific vet, but we also still want to be able to specify a specific vet. Do we maybe need to have them specify Account and Contact here so we can map the person?

We should also make sure this is bringing up the "new" account form that has email included.

deb761:
# Critical Changes

# Changes

Removed Expired and Activation date from form.
Corrected Birthdate so it will show in form.
Added help text in Birthdate.

# Issues Closed
#260

### DIFF
--- a/force-app/main/default/flexipages/DogRecordPage.flexipage-meta.xml
+++ b/force-app/main/default/flexipages/DogRecordPage.flexipage-meta.xml
@@ -68,12 +68,6 @@
                 <identifier>RecordSpayedNeutered__cField</identifier>
             </fieldInstance>
         </itemInstances>
-        <itemInstances>
-            <fieldInstance>
-                <fieldItem>Record.Expired__c</fieldItem>
-                <identifier>RecordExpired__cField</identifier>
-            </fieldInstance>
-        </itemInstances>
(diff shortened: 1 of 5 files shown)


atlasjen:
looks good!

PR #337, opened 2022-09-25
Added classes to AtlasSuperUser

### LINKED ISSUE #319: Add classes to permission set
These need to be added so standard users can access the Atlas1 classes.

Actually, let's create a AtlasUser permission set which is like the AtlasSuperUser permission set but omits the delete all permission on the objects.

Then AtlasSuperUser can add the delete all permissions.

@atlasjen What do you think?

deb761:
# Critical Changes

# Changes

Added description to log object
Created Client list view (might be useful)

# Issues Closed
#319


deb761:
To test, create a standard user in your scratch org, login as that user, and check you can do everything.

atlasjen (changes requested):
Getting this error on deploy to a new scratch org . Likely due to the puppy raiser log PR still being open.
image

PR #338, opened 2022-09-25
320 Client Last Application Received Date not being set in details

deb761:
# Critical Changes

Changed Contact trigger from before update to after update
# Changes

# Issues Closed


atlasjen:
merging. the issue isn't w this field, i realized. will be opening new issue.

Issue #330, opened 2022-09-15
split out Logs for Clients and Team Facilitators


atlasjen:
For client logs, the present Log display is perfect.
For team facilitator logs, add a column for "client", instead of "satisfaction" it should be "client stress".
make sure we can sort by client and date

atlasjen:
I also just realized the comment itself is not displaying. 
so we need to make sure that does display. If we don't show all of the log content in the view in details, then we need something we can click on to see the full details. i do see we can choose edit.

Also update to above if someone is a submitted we need to know who the client is as well.

atlasjen:
is there anything left here? i think we are good.

PR #346, opened 2022-10-05
332 training log updates

### LINKED ISSUE #332: Training Log updates
We need a way to see the full training log. 
Having one of the fields in the Logs related list be clickable would be good.
I believe in later versions you added "view all" capability.

deb761:
# Critical Changes

# Changes
1. Split out logs into two related lists by record type.
1. Cleaned up LogController for add/remove logs (using standard controller for these)
1. Simplified log form a bit:
   1. Removed radio button: user must fill in contact
   2. Removed record type selection: record type is determined by which related list you start at.
1. Added details to the related list: note it is the html contact, so you might see some tags.
2. Removed use of messaging to show/hide dialog.  Passed parameters to the child component for recordtype and recordid.

# Issues Closed
#332, #330


atlasjen (changes requested):
Got this on pushing to a fresh scratch org
image

deb761:
Should be fixed.

atlasjen:
Looking good.  #1 is a fix needed. I am not sure if we can do anything about the others. #2 is less important. for 1.4, fixing #3 is important we can delay on it for this one as i expect it is not simple.

1. The TF log still lists Satisfaction. We should have Client Stress (you have that) and then Stress
image
The "new" TF form looks good other than the "stress" field not being stored/displayed.

2.  [Note, i just saw you warned of this above due to it being a rich text section. ] Are we able to control how formatting comes in to the list view?
For example, I have a carriage return in what i entered for notes
image
and here is how it appears:
image

3. Are we able to control the height of the cells in the list view? I really would like to be able to read across several entries at once. Right now i have to open a log at a time to see the notes.
 
image

deb761:
The simplest solution to 2. above is to strip the html tags.  That means you might loose spaces between paragraphs.  I don't agree with showing these rich text fields in the data table anyway, since the table is a _summary_ of what is in the records, and those fields are detailed information.  I think it makes more sense to have the user view the record so they can see the entire field with formatting.

atlasjen:
I’m fine w stripping html for that view..
But that does directly tie to the later question of if they want to be able to see data from multiple entries at once, what is the best way to do that? Generate a report? It is a pretty reasonable use case to want to glance through and look log to log to see what has changed. I agree not paragraphs, but more than a few words

deb761:
And what if your client or trainer starts out with the same information for every log, and puts differences at the end?  Unless you have a template they are following, you don't have a way to know which part of the field to look at.

atlasjen:
Yeah they have a template they follow. They are always just putting new stuff

deb761:
OK, I'll strip the HTML.

atlasjen:
just need to add displaying "stress" to the TF log
image

looks like you push the html change already.

atlasjen:
This is approved, but looking at the merge, i am not sure if i should just keep the new change or we need the old. please fix the conflict and merge. thank you!

PR #351, opened 2022-10-07
Related PATs changes

### LINKED ISSUE #345: PAT updates
1- need to be able to upload MP4 and MOV
2- need a way to click on it to bring up the full details
3- if access PAT from the easy Nav bar, and choose NEW, there is no ability to upload files
4- if access PAT from the top easy Nav bar, you don't see files that were attached
5- add Files to quick view on right.

the "new PAT" form from PAT

deb761:
# Critical Changes

# Changes
Add link to PAT record page in each row
Remove message channel
Remove extra methods in controller (use standard controller for delete and form submit for create/update)
Problem with create form when add multiple files: the entered data disappears.

# Issues Closed
#345


atlasjen:
Link for forms is now there. thank you! I saw your note and did confirm multiple files upload file.

I am getting an error that MP4's aren't supported, but i think you resolved that in a different PR that isn't on this branch perhaps 
image

I think it will be a page layout change we just have to do in prod, but i don't see the PAT file in the quick view on the right
image

It just has related and doesn't have our custom related file LWC.

deb761:
I don't see a need to use the custom related file component here, since PATs don't have a category or type.

atlasjen:
We had talked about using it so that if we wanted to access these as related files for their contact we could.
But we can punt on that and munge data later if we add it

deb761:
Yes, we had.  But then you would be looking up the related files in Contact for it.

atlasjen:
They’d show up in both places , right? Pat and client? And we just add pat as a client file type. I just didn’t recall if that was fairly automatic or extra work. It is a nice to have.

Get Outlook for iOS<https://aka.ms/o0ukef>

deb761:
Let's add that as a feature for the next release then.  Add PAT as a document type under Client, and when a file is uploaded for a PAT, we add the Client and PAT category/type to it, and create a CDL to the Client for it.

But then it would also be available in the selection for category and type in the file upload dialog.  Not sure we'd want that.  There wouldn't be a way to figure out which PAT to assign it to.

atlasjen:
looks good!

PR #352, opened 2022-10-08
The date was defaulting to today, so the date check failed.

### LINKED ISSUE #339: Last* Date not being displayed if it is today
In testing the various Last* Dates. The common factor of it not displaying in the detail page is if the date is today's date and not sooner. I tried both not selecting a date and leaving it to default to today, and also clicking specifically on today.

I wonder if this got introduced accidentally when you were fixing the timezone display issue.

deb761:
# Critical Changes

# Changes
Removed check for date change on ContentVersion trigger so file service checks every time the record updates.
The date was defaulting to today, so wasn't showing up as a change.

# Issues Closed
#339


atlasjen:
all good!

PR #364, opened 2022-10-15
363 set name for ceu

### LINKED ISSUE #363: Set name for CEU
Salesforce won't let us change that object name to auto-number.  So we need to give these objects appropriate names.  Let's go with [Authority]-[Date], since the length of both of these is known, and title isn't.

deb761:
# Critical Changes
Reverted change to CEU name type

# Changes
Added trigger to set CEU name to CEU-{0000}

# Issues Closed
#363


atlasjen (changes requested):
works great from contact.
From the CEU object itself, it still wants a name set 
image

i would assume it would be much like program assignment that it just sets it in the backend.

deb761:
We can't change the new object layout.  Best we can do is override the new button like we do with Dogs and Contacts.

atlasjen:
Ok. What is it you did w program assignments? That sets the name automatically

Get Outlook for iOS<https://aka.ms/o0ukef>

deb761:
Program Assignment was set up as an auto-number name.  We can't change CEU to that since we released the CEU object type in 1.0.

deb761:
I overloaded the new CEU button to show our new CEU form.  That should solve the initial name problem.

PR #381, opened 2022-11-06
When updating the last date, select all known fields

deb761:
Update python scripts to put all fields in the query. This will prevent added categories or types in production from causing problems.

# Critical Changes

# Changes

# Issues Closed


atlasjen:
nice! looks good.

PR #385, opened 2022-11-15
Fix file sort

deb761:
# Critical Changes

File sort was trying to use an undefined property.

# Changes

# Issues Closed


atlasjen:
that works!

PR #388, opened 2022-11-20
Add hide action to client and facilitator columns

### LINKED ISSUE #365: Allow hiding columns in related logs
We don't have a clear way of knowing which columns the user would like to see in the related logs lists.  So provide a way for the user to hide columns in the list.

Also see if we can change the height of the rows so more lines of text can be seen.

deb761:
# Critical Changes

# Changes
- Add hide to Client and Facilitator columns
- Add button to show all columns next to <new> when a column is hidden

# Issues Closed
#365


atlasjen:
How do i find this/use it?
I am not seeing a hide/show button anywhere either in the direct related list or the all logs list
image
image

Also, we really need to be able to see more from the details on these across multiples, at least on the view all view. Can we make the default # of rows much greater?

deb761:
Oops!  Sorry, it's not particularly obvious:
image

atlasjen:
The hide/show works great.

1) I realized, in this we need a bit more logic in what to show in each table. I stumbled on this as the sample data had some issues we wouldn't have in real data.. namely, in this case a log was labeled "client" but should have been facilitator

Here is example from the sample data (after I set Positions to include Client and Facilitator and gave them active statuses)

image

 none of the sample data has "facilitator" as type

**I would think if it shows in client logs it should only show if the person is the client or submitter.
for facilitator logs, again, it would only show if the person is the faciliator or submitter**

2) and i will now go back on what i said and say let's go ahead and add the submitter column back in each log type w hide show as well. You were right.

3) I also can't edit an existing log to change it's type. Maybe that is by design and that is ok. just meant i couldn't test out if changing that resolved this.

From the toplevel Logs view :
image

It makes sense in your custom view of Client or Facilitator logs that I can't edit the type.

deb761:
1. Are you 100% sure you want to change the logic for which logs are shown for a contact?  I mean, its not as if the contact is viewing the related list.
2. You want to add Submitter to both lists with the option to hide the column.
3. Do you want to be able to change record type?  I took out the select to avoid problems with the user changing types and expecting the record to show up in the table it started in.
4. Do you want me to add Facilitator logs to the sample data?

atlasjen:
1. I want to make sure only the proper log type is shown in the proper section. So, if it is a client log and any of the submitter or client = contact, it should show there. if it is a TF log and any of the submitter or TF= contact, it should show there. I think the main issue is the sample data has blank log types and as a result we are seeing data we shouldn't in the client logs.
2.  Yes. with the logic above
3. No. Let's leave that as is
4. Yes, let's at least make sure all logs have a type and are consistent with the logic above

atlasjen (changes requested):
see comments

deb761:
Fixed data and added Submitter.

atlasjen:
perfect!

atlasjen:
confirming. closing. thank you.

PR #396, opened 2022-11-30
Show N/A for health status if dog status is None

deb761:
# Critical Changes

# Changes
Modified Dog Health Care Record Status to show N/A if the dog status is None
# Issues Closed


atlasjen:
looks pretty good. I like the icon you came up with. the only other place to change for consistency is the summary
image

We should also have it be N/A if NO status is set as well as "none". i noticed if i blank the status it doesn't show N/A.
image

atlasjen (changes requested):
see comment

atlasjen:
perfect.

PR #408, opened 2022-12-17
Changed trigger from Team to PAT

### LINKED ISSUE #394: update trigger on PAT Valid Until
we presently have a trigger on Team. PAT Valid Until that sets Contact. Client Certification Valid until.
However, the Team. PAT valid until has to be set manually and is not connected to the individual PAT.valid until.

Please change it so the thing we set is the PAT. Valid until. Then, the Team and Contact ones should update to have the most recent date of any PATs for that client. 

I had gone through and fixed some of the valid until dates on the PATs but then realized i had to change them again on the team.

deb761:
# Critical Changes

# Changes
- Changed Trigger to be on PAT rather than Team
- Trigger updates related Team and Client

# Issues Closed
#394


atlasjen (approved):
looks good!

PR #412, opened 2022-12-26
Add trigger to Team to create dog relationship

### LINKED ISSUE #403: Dogs under contact vs Team- setting dog reln on contact
I was explaining teams vs clients and dogs to molly and niki tonight.
With the plan to teams driving more and more (status, seeing files), it would seem that if i create a team, then if the dog is not associated w/ the contact, it should add the dog relationship. Right now it does not and I have to manually add the dog to the contact as well.

I don't think we add logic of if I add a dog as "client" that it creates a team. Dog is just hanging out.
Similarly, if team is deleted, I don't think we delete the dog relationship if it exists.

deb761:
# Critical Changes

# Changes

Created trigger on Team create and update to create a Client dog relationship between the Client and Dog.

# Issues Closed
#403

### DIFF
--- a/force-app/main/default/classes/DogController.cls
+++ b/force-app/main/default/classes/DogController.cls
@@ -5,10 +5,7 @@ public inherited sharing class DogController {
             return [
                 SELECT
                     Id,
-                    VaccineDueDistemper__c,
-                    VaccineDueHepatitis__c,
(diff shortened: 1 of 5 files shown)


atlasjen:
looks great!. merging

Issue #405, opened 2022-12-13
look at having Teams drive last* dates for client and dog records


atlasjen:
ideally want to have access to all "last*" dates on client and dog records in Teams.
It would possibly make the most sense to set them on the team and have a trigger to copy them to contact.

contemplate best way to do this.

atlasjen:
I spoke with Niki and Molly.
Here is the thought, and this aligns with #404 and #395 and #400 in consolidating data to teams.
use case flow is going to likely be
- someone (programmatic or manual) creates a contact. From there they create a team (which has the step of creating a dog along the way)
- As files come in that are needed to apply, those all get added to the team, be they for the client or dog.
- when it comes to client annual recerts forms those get set on the team (PATs are already on the team)
- when it comes to dog vaccinations, those stay w the dog. we have all of that fancy logic there.
- I am envisioning an LWC on the related tab of Team that has these Last* dates set by uploading files in the related files component in teams (see #400)

From the contact side, here is what i think moves to being set in the team and needs to be displayed in the team
![image](https://user-images.githubusercontent.com/66139437/207483241-25eb8df9-1af3-43d5-9d65-839f206de424.png)

If it has a "C" it stays on the contact side and is just associated w the contact, not a team
If it has a "T" it isn't displayed on the contact at all. It really should be a new team field and we deprecate these
If it has a T->C, it is set on the team but displayed in the contact.

From the Dog side, here is what i think moves to being set in the team and needs to be displayed in the team
 
![image](https://user-images.githubusercontent.com/66139437/207483707-94bb3a93-f656-423d-9ec3-58468ad48f7a.png)

If it has  a "T->D" it is set on the team and copied to the dog
If it has a "D" it is just on the dog. all of the other stuff on dog is just on the dog.

- Some considerations. IF someone edits data on a team, it should stomp on whatever is on the contact or dog.
- People will still upload some files to Contact and some to Dog. but we will use business process to say what gets uploaded where.
- Ideally we would have (info) little icons by any field that is set by a trigger to say don't touch. This also includes the Team PAT Valid Until which is set from the PAT

deb761:
I could put a form to update the client and dog data in the team details, or do you also need the status in the Team?  Alternatively, we could have a form in the client that sets the team fields.

atlasjen:
I think we set status on the team and it pushes to client and dog. I need to be able to have status on teams.

deb761:
[...]

Issue #400, opened 2022-12-13
Add related files to teams


atlasjen:
We have a use case of needing to see the files we have for a team all in one place.
1. If a contact has client category files uploaded, and they are in a team, show the client files in the team.
2. if a dog had dog category files uploaded and they are in a team, show the dog files in a team
3. allow the ability to upload client AND dog records from within a team. the dog records should get assigned to the dog and client to the contact.
![image](https://user-images.githubusercontent.com/66139437/207206764-1eccb490-2c1d-4aac-907c-d7c2366519bb.png)

atlasjen:
A little confused on expected behavior.
I made a team in the scratch org.
There are now 3 related file sections
![image](https://user-images.githubusercontent.com/66139437/209057722-662fa546-7098-4461-8dc3-904724483583.png)

Under Teams, Related:
If I upload things to Files (which i can pick Client or Dog as types,_ they do not show in the contact or dog.
If I upload things to Client Files, they do indeed show in Contact.related files.

Under Contact. Related:
If I upload a client category file, it does appear under Teams . related  Client Files
Under Dogs. Related:
If I upload a dog category file it does appear under Teams . related  DogFiles

However, these do NOT appear under Teams. Files.
And if i upload a Client or Dog category file under Teams. files, they don't show in the Contact or Dog. or under Teams.Contact Files or Teams.Dog files in the related tab

It seems ideally we'd just use the Related Files LWC under Teams and filter based on the category (dog or client) vs adding in the 2 addl Client Files and Dog Files sections.

---
I am also able to pick any category i want (trainer, etc) in the Client Files and Dog Files sections under Teams. Related. If we keep these, it seems they should filter to only be their type.. but again, I'd be inclined to say they don't need to be there and we just use Related Files and 'category' under Teams.

---
know issue, but a bit trickier w/ this logic, i can set Dog files on my Contact and Contact files on my dog. If I set a "dog" category file on my contact it shows in BOTH the teams dog and client sections... which does make sense given the logic.
I'd say best is to not display "DOG" as a category in the contact context and not display anything other than "DOG" as a category in the dog context

atlasjen:
I also got this error pushing sample data deploying this to a fresh scratch org:
                    Prepared 80 rows for DataOperationType.INSERT to Log__c.
[12/21/22 20:16:10] Running step: Insert ProgramAssignment__c
[...]

PR #413, opened 2022-12-29
395 405 team status

### LINKED ISSUE #395: Teams/Client/Dog Status
In working on a report of how many teams we've ever certified, i realized i can't pull that form a contact's client status as that is their current status.. i need to be able to look at the team... but there is no status.

Here is my suggestion

- we set status on team manually
- there is a trigger to set it on client and dog to be the same thing. If there was already a status on the dog or client, we overwrite it w/ what is in team
- if someone happens to manualy edit the client/dog status later, oh well.  The next time it is updated in the team, we stomp on it.

this is actually staying consistent w/ how we derive certification valid until on the client from the team.

deb761:
# Critical Changes

# Changes

1. Moved LastClient*Dates to Team.
2. Added Team status.
3. Added linking these files also to the Client and dog.
4. Removed LastClient* Dates from Contact layout and added them to Team layout.
5. Added linking dog files to all teams they are in.
6. Added copying Team status to ClientStatus on the client contact along with CertificationValidUntil.

# Issues Closed
#395 #400 #405


deb761:
Note: I'm not done yet, just wanted to continue the conversation.  I don't have any tests in that check the CDL is created for the contact or dog.  I thought I did, but I don't see them anywhere.

deb761:
This is ready for review now.  Sorry about the delay.  I'm in water heater/repipe limbo.

atlasjen:
Good deal. I am out of town at a conference until Friday. I will see if I have breaks and can review before then
Good luck on the piumbing!

atlasjen:
1. I thought we made this 2 way so that if files were added on client or dog, they would also show on team and vice versa. Maybe I am incorrect. Maybe it was just dog that was to be 2 way.
*I tested the use case of if client files were added to the client before the team existed. I would have expected them to also show in the team, as well as the "last date" once the team was created.
*I also tested adding "client" files to the client once the team existed and they didn't show in the team, nor did the "last date"
*I also tested adding "dog" files to the client once the team existed and they didn't show in the team, nor did the "last date"
I do know we were not going to have status be 2 way and that was just from the team and would overwrite anything in client/dog..
I am almost positive we talked about this one being 2 way for the dog since loading the dog files will also be the time they set the vet health info, so they'd be in the dog object.

The trick w/ this is we'd have to always make sure it was the right client/dog being updated. this isn't a problem from team->client/dog but from dog/client->team could be. we'd have to be careful it is going to the proper team, not just any team that client was in. thinking about that, not sure how we'd do that from client since they will still have the relationship to the old dog and old team.  From the dog, my recollection is we said dog records are dog records and would apply to any team, thus ok to overwrite.

3. Dog status doesn't update when team status does.
I updated the status in team. It correctly updated in the client but not the dog.

4. were we going to display the dog health status block in the related list in the team somewhere? The one rollup field "health record status" would be fine.

5. It looks like "certification valid until" doens't populate on the team when the PAT value is set. Realizing in our model certification valid until is the exact same as PAT valid until.. so since we are no longer showing it on the client, that is redundant. We can drop "certification valid until"

6. We should display Client and Team Facilitator Logs related LWCs in the Team as well.

some other tests that worked as expected
* Files added to team show up in client or dog w correct dates as expected
* contact status updates when team status is updated
* loading PATs  sets PAT Valid Until in the team.

deb761:
1. I only made it one-way, since contact-team and dog-team is a 1 to many, we can't be sure which team to attach a document to.
2. Yep.  Looks like I missed updating the dog status.
3. I was going to add dog health in another PR.
4. The logs are in a different PR.
5. In Issue #395, you asked for the trigger after manually setting team status.  It is _very_ frustrating when you change the acceptance criteria.

atlasjen:
1. Agreed 1 way makes sense on the client but we had talked about for the dog anything on the dog would apply to any team since it is health info. . since they have to manually set the vaccination info at the same time as uploading the files, having to do that in 2 locations is going to be really inefficient. open to other ideas if we want to move some of the dog stuff into team.

sorry on 3,4,6. 

5. I'm confused the issue you referenced just refers to status which is indeed setting manually on the team then sets it on dog and client. Nothing changed there.  I was bringing up realizing we didn't need to move Certification Valid Until to the team as it is redundant w/ the PAT valid value which does already get set.

deb761:
So in 1 you are saying the dog status should overwrite the team status _AND_ the team status overwrites the dog status?

atlasjen:
This is likely an easier one to get on a call to go over. Let me know if you have any time this weekend. I am doing a lot of various sessions w clients but can def make time

PR #415, opened 2023-01-25
331 view all files updates

### LINKED ISSUE #331: View all Files updates
Since we are using custom component to 
view all files, logs etc,
add the ability to filter not just sort.
Filtering by Type will be the most important.

deb761:
# Critical Changes

# Changes

Added filter button and select filter dialog to related files component.

# Issues Closed
#331


atlasjen (changes requested):
The filtering works great. I think an unintended side effect of not showing "filter" if there are no files is it now no longer dynamically refreshes to show when the very 1st file is uploaded.

Here is what i tried
- in this scratch org, found a contact who had no related files. I uploaded a file. I got a confirmation it uploaded, but until I refreshed the page it didn't show.
- in this scratch org, once any one file was shown others dynamically refresh
-in prod i tested the same use case and it dynamically refreshed on the 1st file.

Otherwise, filtering is great! 

Is there a way to  somehow visually show that filtering had been applied?
Esp when in the view all files view, you have no way to know if something is filtered and other files are there. like display what it is filtered by or change the color of the filter button if it is active? Just someway to know.

atlasjen:
merging. will add as a future feature request to see in the view if a filter has been applied (vs needing to click on filter)

PR #427, opened 2023-02-09
Move Contact position details into actions

deb761:
# Critical Changes

# Changes

1. Moved Contact details related to each position into a action layouts so we can hide those not relevant to a contact.

1. Added Is... formula fields so we can hide position details based on positions assigned. 
2. Moved position details into tab on contact record page.

# Issues Closed


atlasjen:
OOH I like this! nice to have less on the main page.

PR #432, opened 2023-02-11
Add TA Requested field to Program Assignment

deb761:
# Critical Changes

# Changes

- Add TA Requested field to Program Assignment
- And add to related list in contact

# Issues Closed
422


atlasjen (changes requested):
i think this needs to get added to the admin.profile-meta.xml.
I see that the program assignment object and page layout defn have been updated, but i don't see them in
- Program Assignment List view
- Add/Edit Program Assignment

deb761:
Added to permission set and list views.

PR #434, opened 2023-02-16
430 add dog as a column in tf and client logs

### LINKED ISSUE #430: Add "dog" as a column in TF and Client logs
I am not sure if this is something we do here or in prod, but i want "dog" to also display as a column when we look at TF or client logs. Right now it does not. This should be included pretty early in the columns so it is right after the client and handler names. This would apply to viewing it under contact or team.

Right now our "team" filtering on logs just seems to filter on client, not also dog and we don't have any logic that forces logs to be related to valid teams. As a first step, let's include the dog name on display. it is a later issue (diff release) to address the logic differently.

deb761:
# Critical Changes

# Changes
1. Modified LogController to check object type from recordId and return only the logs for the dog and client for teams.
2. Fixed the breadcrumbs in view all related items.
3. Removed the relatedTeamLogs LWC as no longer needed.

# Issues Closed
#430


atlasjen:
I tested a few use cases.
- WORKS: Client logs properly show for just the right combo of dog and team under Team Related logs
- NEEDS CHANGE: I added a TF log for the client and dog in a team and it does not show under the client/dog team related TF logs.
- NEEDS CHANGE:  While we don't need dog as a column in the team related logs, we do need it under Contact since if someone is in 2 teams, there is no way to tell which dog the logs are about. This would apply to the TF and Client logs.. unless you want to display/link to the teamname instead.. see below for addl thought on this.
- THOUGHT: I wonder to make this simpler if rather than entering the client and dog separately on NEW log forms, on a client log they pick a team and a TF log they pick a team.. that way the only way logs can be filed are to valid teams.  We could still display "client"  and "dog" in the record under contact, but in the one under the Team we wouldn't have to display that if we wanted to save space. I'm not sure if the component in both is the same. If it is , then just add "dog" for display OR drop client and dog and use the teamname w a link.

deb761:
1. I think it makes sense to link the logs to teams rather than matching client and dog.  We can still show the Client, TF, and dog in the list if needed, but showing the Team name with a link is about the same thing as Client and Dog.
2. The bug with the TF log is not expected.  I'll look into it.

atlasjen:
1. TF logs look like they are populating well now, thanks!
2. Let's update the Logs "all" and "recent" views to include Team 
image
3. The "new" and "edit" buttons on Logs (when viewed from logs list view, not team) needs to be overloaded to bring up the form that uses team vs client and dog
image
4. you will want to update the sample data for logs to add team 
5. Facilitator should be required on the TF Logs (both if set from the contact, team or "logs")

deb761:
1. I had updated the sample data, but you must have used skipped loading it.
2. How about we replace Client and Dog in those list views with Team?
3. How about we hide the Client and Dog fields?  Would the default new dialog work then?  We would definitely need a custom form if Facilitator is required in TF logs but not on Client logs.

atlasjen:
1. I did a pull and then pushed so maybe that skipped loading data ; I bet it did 
2. Yes team only works 
3. Ah if it requires a diff form them just leave it optional. Tf’s should be smart enough to put their name in as it is their log.

deb761:
I've updated the list views.  Need to update the PostInstall script to populate the new Team field for each log.  Overriding the new button for log is a lot of work, much better to use the default form for now.

atlasjen:
Pulled the latest to a new scratch org.
1. The "log" NEW form still uses client and dog vs team (for both client and tf logs). if accessed via Logs. If accessed via contacts/teams, it is fine.
image
image
3. I am getting an error on the data load
 [02/20/23 16:39:00] Prepared 10 rows for DataOperationType.INSERT to ContinuingEducationUnit__c.
                    Running step: Insert npe4__Relationship__c
                    Creating DataOperationType.INSERT Operation for npe4__Relationship__c using smart
                    Prepared 0 rows for DataOperationType.INSERT to npe4__Relationship__c.
                    Running step: Insert DogRelationship__c
                    Creating DataOperationType.INSERT Operation for DogRelationship__c using smart
                    Prepared 50 rows for DataOperationType.INSERT to DogRelationship__c.
[02/20/23 16:39:01] Running step: Insert Log__c
                    Extracting Record Types for Log__c
                    Exception in task config_dev.load_dataset

Error: 'Team__c_sf_ids'

Everything else is looking great !

deb761:
I think the problem with the default New dialog is that the dev_org flow updates the Admin profile to have access to everything.  If you modify the Admin profile to remove access to Log.Client and Log.Dog, they won't be in the new dialog.

image

atlasjen:
Agree on pulling handler from Log as, you are right, they are part of team. Let's drop them from the log.

I changed the admin profile in my scratch org. I think one thing that confused me is team was all the way on the bottom of the form and i didn't even see it. Can we move it to the top? Maybe right above handler? The new client and new facilitator forms you get from w/in the related lists are great. Team is up top. So if we could have them be consistent when accessed through Log that would rock.

Everything else tests out, so all that is left is move Team up higher in the New Log form that is accessed from Log. Woot!!

deb761:
Sadly, we can't change the default new dialog.  How important is it to move the Team item?

atlasjen:
I can get people used to going down there, it’s just super non-intuitive..

deb761:
Thankfully, I was able to improve those dialogs:
image
image

atlasjen:
looks good to me!

PR #447, opened 2023-03-22
139 change file upload to be contextual

### LINKED ISSUE #139: change file upload to be contextual
If in contact, only show "categories" based on positions associated with that contact.
If in dog, only show dog category.

We do not need to block first release for this.

deb761:
This is a rather large change.  It required adding an object to define how the context would work, and a controller that would provide the desired list of categories for a given object.  Note that because a new object type is added, you will want to load_dataset in the org you are testing this in, which is run as part of the dev_org flow.

# Critical Changes

1. Made Categories a global picklist so they can be shared with CategoryRules.
2. Added CategoryRule object and sample data.
3. Created CategoryRuleController that would return the list of categories for an object.  For Contacts, Positions are part of the rule.
4. Created FileInformationCmp that uses the controller to get the list of categories for an object.
5. FileInformationCmp provides selected Category, Type, and Date values to parent containers.
6. FileInformationCmp gets an existing ContentVersion information to show current category, type, and date.  Provides this information in an event to parent container.
7. FileUploadCmp and FileUpdateCmp now also need the recordId of the object the file can be/is linked to.

# Changes

# Issues Closed
#139


atlasjen:
I pushed this to the same org i did the other 2 PR's in. Let me know if that is an issue and if i should do a fresh deploy. I didn't get an error.

I opened a contact "Robyn Bean" and went to the related tab. I clicked on Files to upload but i am unable to get anything to show under category or type. This test data had Staff as a position. I added volunteer just in case it was based on position.. but no change.. There are no errors, it just doesn't let me see any options. 
image

atlasjen:
Testing results- think the only 
* **Good** I looked over the mapping in the controller. that looks right. makes sense that ADSiM and Standalone will appear for any position
* **Good** Teams file upload works great. Both client and dog work perfectly. I double checked nothing broke w the logic of being able to see it on the contact or dog.  Files also properly show on the Contact when uploaded even if their position isn't set to Client (dummy data, won't be an issue in prod)
* **BUG** Dog- i am unable to select any types (position is properly Dog). Should have option of "Vaccination and Exams"
* **Question/Good** Contact- For a position that is already set, i am able to upload files. However it looks like a page reload  on the related tab is necessary in order for the new category to appear. I am guessing this is expected behavior, but just double checking.  I also see that once a file is uploaded if a position is removed the file remains, so that is also good.

So, assuming the refresh is to be expected, which we can just document as known behavior, then it's just fixing dog. super close!!

deb761:
1. Dog problem was a timing issue.  Should be fixed.
2. Changing the positions would require a page reload for the category list to change.

atlasjen:
Dog is fixed. :)

So I just realized in testing dog that we should show different things when you are in the Team view to upload dog or client files than in Contact or Dog. This will help avoid people uploading them to the wrong place, which is a big part of the point of this change. it's less of an issue if they get added to team as they can be seen in client/dog; it's the otherway around that is more of a problem. Here is how they'd be split out (same as the page layout)

**Team.Dog-**> Vet Health Form, Vet Xray Report
**Dog.Dog-**> Vaccinations and Exams

**Team.Client**->Pre-Application, Application, LOMI, Veteran Paperwork, Application Video, CGC
**Contact.Client**->Contact Form

One question on this, though. right now one of the fixes we had in 1.4 was we could support adhoc types as long as we didn't have dates associated w them. how will that play into this? The place it is going to come up the most is on Contact.  Can we have it be that these types are explicitly filtered to only show as above and anything else will show per the Category/Type mapping under ContentType?

PR #451, opened 2023-03-31
410 background check types

deb761:
# Critical Changes

# Changes

1. Added Type to BackgroundCheck object, with values of 'Criminal' and 'Credit'.  Default is 'Criminal'.
2. Added Type to related list and search list.
3. Added PostInstall to set all BackgroundCheck types to 'Criminal'.
4. Updated BackgroundCheck form to include Type.
5. Updated sample data.

# Issues Closed


atlasjen:
Looks good. If it is easy to set all types to Criminal that are in the system in prod, that would be great. I'll manually fix the few that are credit. If that is a pain , it's not that many and we can manually set them.

PR #457, opened 2023-04-12
454 filter team

### LINKED ISSUE #454: more file upload updates- filter by type
So I just realized in testing dog that we should show different things when you are in the Team view to upload dog or client files than in Contact or Dog. This will help avoid people uploading them to the wrong place, which is a big part of the point of this change. it's less of an issue if they get added to team as they can be seen in client/dog; it's the otherway around that is more of a problem. Here is how they'd be split out (same as the page layout)

This has become more important as we have defined that we only want certain files uploaded in certain locations.
Under Team, they should only be able to upload Client and Dog files associated w team.
Under Contact, they should not be able to upload Client and Dog files that are supposed to be on team.
Under Dog, they should only be able to upload dog files not on team

Team.Dog-> Vet Health Form, Vet Xray Report
Dog.Dog-> Vaccinations and Exams

Team.Client->Pre-Application, Application, LOMI, Veteran Paperwork, Application Video, CGC
Contact.Client->Contact Form

One question on this, though. right now one of the fixes we had in 1.4 was we could support adhoc types as long as we didn't have dates associated w them. how will that play into this? The place it is going to come up the most is on Contact. Can we have it be that these types are explicitly filtered to only show as above and anything else will show per the Category/Type mapping under ContentType?

deb761:
# Critical Changes

# Changes
- Added N/A category for ContactForm document type
- Removed ContactForm from the other categories
- Added Team Dog category for HealthForm and XRayReport document types
- Added custom setting to turn off linking Team files to Client and Dog.  This is used during PostInstall and tests to avoid "Too many SOQL" query errors.
- Updated FileService for new category.

# Issues Closed
#454


atlasjen (changes requested):
I tried opening the Upload File Dialog under Team, Contact and Dog and it silently does nothing when i try to pick "Category"

Not sure if related, but in this scratch org (new org w/ this push) Leads no longer are showing Related Files.

I see that there are no sample "Category Rules" so that may be the issue.
Also not sure if related, but this message was in terminal
image

atlasjen:
Testing this most recent version
- Team Dog and Team Client files are not visible in Client and Dog. I see your comment above about removing that link... but we def still want to be able to view them under Contact and Dog. Type would just be "team client" or "team dog".
- from the rest of my testing file uploads and dates in contact and dog all seem to work as expected
- using arbitrary new Document Types seems to work. I couldn't try it in client due to the next item
- "Client" is now missing as a Category in the Document Type mapping under Content Version
image

Other seemingly related file upload things
- File upload dialog under Background Check shouldn't be this custom dialog; it should be normal system one [CEUs seem to use the proper one, so not sure why background check is different]
- file upload still missing from Lead (i assume a page layout thing)

deb761:
I renamed Client to "Team Client".  N/A only has "Contact Form" in it.
I haven't touched the Lead record page at ALL for this PR.  Not sure why Files isn't in there.  I'll check it out, maybe there is an LWC error.
Ah, there are no Category Rules for Lead.  Should there be?
As for the Background Check record page, it has the custom Related files list, or are you talking something else?

atlasjen:
Oh right, i was thinking Client would still have the ADSiM files, but they have their own category but display under client details. I would say it might make more sense to rename Assistance Dogs Set in Motion as a category to "Client" and then have file types of ADSiM Pre-Application, ADSiM Application, ADSiM Program Agreement.  The other reason for this is if there are other random general files we want to upload under client, but that don't have last* dates that gives us a way to do that.

For background check, it should just have the normal file component. That is what it has in prod right now.

We could overload the File Upload for Lead. Right now it is the normal file upload component and we classify the files when we convert to contact. Typically they will be volunteer resumes or interview notes, but as a lead can go any which way and won't have many files, i figured we just leave it the generic and they become files on contact when converted. I think we leave this alone.

Then the only other issue in my list above is that Team Client and Team Dog records would still be copied to display in the Contact and Dog locations like they did previously.

atlasjen:
Agreed- leave Assistance Dogs Set in Motion as a category.
Leave Client out. If start wanting arbitrary client files, can manually add to Content Version and Category Rules w/o requiring code changes as long as don't need  "last * dates"

deb761:
I changed the label for Client category to Team client. Since I thought you said you only wanted the client options available for teams. Since these Sent from my iPhoneOn Apr 15, 2023, at 4:25 PM, atlasjen ***@***.***> wrote:﻿
Testing this most recent version

Team Dog and Team Client files are not visible in Client and Dog. I see your comment above about removing that link... but we def still want to be able to view them under Contact and Dog. Type would just be "team client" or "team dog".
from the rest of my testing file uploads and dates in contact and dog all seem to work as expected
using arbitrary new Document Types seems to work. I couldn't try it in client due to the next item
"Client" is now missing as a Category in the Document Type mapping under Content Version


Other seemingly related file upload things

File upload dialog under Background Check shouldn't be this custom dialog; it should be normal system one [CEUs seem to use the proper one, so not sure why background check is different]
file upload still missing from Lead (i assume a page layout thing)

atlasjen:
relates to #478

PR #459, opened 2023-04-12
Propagate Team status when not active

### LINKED ISSUE #453: "discontinued" and "decertified" team statuses not propegating to contact or dog
This is on the 1.4 release. Molly just caught this and I can reproduce it.
If you set Team status to either Decertified or Discontinued it does not update the contact client status or the dog status.
If you set it to other values like None or certified it does propegate.

I tried going in and changing to something else and back to these, but no luck. All values appear to be in the picklists correctly.

deb761:
# Critical Changes

# Changes
- Modified Team trigger to propagate status even when Team not active
- Updated tests

# Issues Closed
#453

### DIFF
--- a/force-app/main/default/classes/TeamService.cls
+++ b/force-app/main/default/classes/TeamService.cls
@@ -32,15 +32,9 @@ public with sharing class TeamService {
             update relationShip;
         }
     }
-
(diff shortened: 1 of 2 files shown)


atlasjen:
all good.

PR #469, opened 2023-04-20
Cancel buttons now redirect as expected

### LINKED ISSUE #462: Error when canceling out of New Background Check
If i go to make a New Background check, and then choose Cancel, i get this error.
image

deb761:
# Critical Changes

# Changes
Fixed Cancel on New background check from related list.
Also fixed cancel on BackgroundCheck and CEU new button overrides.

# Issues Closed
#462

### DIFF
--- a/force-app/main/default/lwc/ceuFormCmp/ceuFormCmp.html
+++ b/force-app/main/default/lwc/ceuFormCmp/ceuFormCmp.html
@@ -3,6 +3,7 @@
         title={title}
         submitlabel="Save"
         onsubmit={handleSubmit}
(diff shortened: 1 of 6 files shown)


atlasjen:
I don't get an error any more, but if i hit cancel on a new CEU, it does not bring me back to the CEU listing.
I just have a blank page
image

If I hit cancel on a new background check, i am returned to the background check listings.

deb761:
I'm not sure what was going on with your org, but it works for me.

atlasjen:
ok. i will merge and we can see how it is then. maybe a scratch org fluke

PR #472, opened 2023-05-07
Remove setting home email from email

deb761:
# Critical Changes

# Changes
Remove setting home email from email in trigger

# Issues Closed


atlasjen:
looks good. realizing w/ workflow rules we actually can get the same behavior anyway

PR #479, opened 2023-06-17
Remove default values for category and type in lwc

### LINKED ISSUE #477: Error when select Upload Files on Contact
testing in staging, I tried contact "fred test" and went to related files and chose upload.
image

I have a feeling it is due to "Board member" or "Staff/Contractor" being one of the positions. If either position is set, this error happens. If they are not, it works.

update: also failed if "Team Facilitator" is set as a position. Does work if "client" is set.

deb761:
Add CategoryRule data to beta dataset

# Critical Changes
Fixed null reference exception when uploading files using the component with no document category and type defined.

# Changes

# Issues Closed
#477

### DIFF
--- a/datasets/beta/beta.dataset.sql
+++ b/datasets/beta/beta.dataset.sql
@@ -1,0 +1,22 @@
+BEGIN TRANSACTION;
+CREATE TABLE "CategoryRule__c" (
+	id INTEGER NOT NULL, 
+	"Category__c" VARCHAR(255), 
+	"Object__c" VARCHAR(255), 
+	"Position__c" VARCHAR(255), 
+	PRIMARY KEY (id)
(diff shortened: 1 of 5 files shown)


atlasjen:
I tried going from Theodore Hale's contact record and chose to upload w/o category or type specified and get this error
image

atlasjen:
Good news is i no longer get an error when i pick Staff/Contractor as type and am able to upload those, so must be related in some way to the same fix.

deb761:
I deployed a fresh scratch org, and added a file with no category or type to Theodore Hale without getting that error.  I'm not sure how you got it.

atlasjen:
i had the wrong branch checked out. works for me.

PR #488, opened 2023-09-19
Group access

deb761:
# Critical Changes

# Changes
1. Added groups to control file access
2. Added logic to FileService to check groups for related files
3. Added logic to CategoryRuleController to check groups for category access

# Issues Closed


atlasjen:
Here is what i did so far to test
- created a new user (atlas min user profile)
- added them to public group "volunteer" 
- logged in as them
-clicked on related and got 2 errors
1) You do not have access to the Apex class named 'CategoryRuleController'.
- went to the Volunteer Admin User
- Apex Class Access, CategoryRuleController is enabled.
2) undefined error   

then, 
- went to update permission set for Volunteer Admin User, but it doesn't let me add a user

image

atlasjen:
tested out adding users to other permission sets  also fails. i can only add them to the permission set i am in (atlas super user)

atlasjen:
- deployed the latest
- created test user w/ Atlas Minimum Access profile "testuser@atlasdog.org"
- ran <code>sfdx force:user:permset:assign -n VolunteerUser -o testuser@atlasdog.org</code>
got this error
image

when i look at the permission sets, i see "Atlas Volunteer"
image

-reran <code>sfdx force:user:permset:assign -n AtlasVolunteer -o testuser@atlasdog.org</code>
that was successful

- It seems to be doing what is expected. If the individual's position contains volunteer, i can see Position Details for volunteer. I can't see any other position details. I also don't see background checks, teams or dogs and get an error if i try to access them (this is good).

I know you are still working on filtering the file views.
I think ideally we would also hide any PII (email, phone, address) for anyone who didn't have volunteer as a position.
I'm going to commit this but to do
- fix readme
- hide PII (i think you said there is some built in salesforce option for this)

PR #490, opened 2023-10-01
Create permission sets

deb761:
# Critical Changes

# Changes
Added several permission sets, including Atlas Base

# Issues Closed


atlasjen:
I first tried just updating my existing scratch org and got this error, so i started fresh.
image


I tried running this on a new scratch org and got this error. Note, i realized after the fact that i also got this API warning when i tried to do a push up above.
image

deb761:
Try updating your cumulus ci.  pipx upgrade cumulusci

atlasjen:
i updated it.. still the same issue. there are two things here, 

1 is the error adding the sample data.
2 is the cumulus error

atlasjen (changes requested):
see comment below. i have the latest cumulus version. I am not seeing this error w other PRs (like the one where you added initial certification dates) just this one.

atlasjen:
ran successfully this time

atlasjen:
- Background check does what is expected.
- Demographic does what is expected.
- Tested the Team permission set.

It adds Client permissions, which is expected but it did not add dog permissions.
I get this error when i try to open the related dog.
image

I would expect Team to also include Client and Dog.

PR #494, opened 2023-10-03
Add Initial Certification dates

### LINKED ISSUE #484: Add three more date fields for positions
We have realized we need to be able to save these and use them independent of when they sign the agreements, as those often lag.

**We need one on Team:**

- Initial Certification Date

It can get added to the right column on team under status
image

**We need these on Contact:**

- Client Initial Certification Date [ derived from Team. Initial Certification Date, so add tool tip consistent w/ others to not set directly
- Trainer Initial Certification Date
- Facilitator Initial Certification Date

These would go under Position Details in the respective section.
Let's make them the 2nd field on the left column of the page layout, so in between the status and Valid until fields
These are all optional.

image

deb761:
# Critical Changes

# Changes
Added Initial Certification Dates for Team, Client, Facilitator, Trainer
Add copy of Team Initial Certification to Client
Add to test
Add to Actions (will need to do this manually in production)

# Issues Closed
#484 #497


atlasjen:
I see these in the Contact and Team objects when i look at fields and i see it in the Team page layout, but they do not display in the  position details pane on the Contacts. I'm not sure how to edit those layouts.

image

atlasjen:
the fields are present. They need to be added to the page layouts.

deb761:
The fields should be in the details pane now, along with updated logic for the Valid Until field on Trainer and Facilitator.

atlasjen:
i played around w the logic a bit.
It appears: Valid Until is Initial Cert Date + 1 year. that is correct.

Then, if i update Last Certification Agreement to a date AFTER initial certification, it correctly updates the valid until to the next year on the original date. that is also correct.

 However, if i update Last Certification Agreement to 1 day before the Valid Until, the Valid until doesn't change.  

I tested this on facilitator, but i assume it is the same on trainer.

logic gets tricky here as , i can totally see the case of we get their certification agreement a few days after we set the initial cert date and we wouldn't want it to roll the valid until forward. How about this- if last certification agreement is w/in 90 days of when valid until expires, we roll forward valid until?

atlasjen (changes requested):
see earlier comment on dealing w/ date that is slightly ahead of the last agreement date (up to 90 days).

deb761:
I suggest 30 days before the certification expiration, 90 days seems a bit big, but I doubt it would hurt.

atlasjen:
i am fine w/ 30. in asking Molly, apparently present business process is they are sent request to renew 30 days before they are set to expire

atlasjen:
great!

PR #495, opened 2023-10-07
485 add standalone position

### LINKED ISSUE #485: Add "Standalone" as a position
Right now it is really hard to tell who is active in something or not in a way we can do bulk emails.
Rather than having the position be blank, let's add "Standalone" or "Standalone Student".

This would cover any of the standalone courses including TSiM. We are pushing these standalone courses more and more and we need a good way to access info about them in Contacts.

Statuses would be similar to trainer and others: [Standalone Status]
Initial Inquiry
Onboarding
In Training
Inactive
On Hold
Discontinued
None

For the File upload component, we already have Standalone Programs as a category and we already have these two file types. so no change there other than setting the last* dates.
- Application
- Program Agreement

For the Standalone position details we would have these set based on the file uploads.
- Last Standalone Application
- Last Standalone Program Agreement

Then there are these related updates (not sure if i missed any)

- add Contact.IsStandalone and code to populate it
- update Category Rule to add Standalone as a position for Contact
- update code to display some sort of icon in the Position(s) with Status detail pane

We can do an import after the fact to set standalone status, and position as needed. I can manage that.

Are you able to programmatically backfill the dates for the standalone application and agreements based on what has been uploaded to files?

deb761:
# Critical Changes

# Changes
1. Moved Standalone into a position
2. Created Last Application and Last Program Agreement fields
3. Added these fields to a quick action that shows on the Position details tab for the Contact layout
4. Update file upload trigger to set these fields
5. Updated Contact Status component to show the Standalone status rather than check program assignments

# Issues Closed
#485


atlasjen:
looks great.

PR #496, opened 2023-10-08
Update board term valid until logic

deb761:
To last day of the month 3 years after the start date

# Critical Changes

# Changes

Change calculation for board term valid until logic to last day of month 3 years after the start date.

# Issues Closed


deb761:
Question: do you want me to change the date only when the last board term start changes?

atlasjen:
well.. we have continued bylaw edits, so apparently it is decided the logic is going back to the way it was ... ish other than our annual meetings are now going to be in April.. so we need to
- revert to the previous logic
- change it to add however many months to get to end of april vs end of june
sorry!

atlasjen:
looks good.

PR #500, opened 2023-10-20
Added the checkbox Text Messages Approved

### LINKED ISSUE #498: do we add a checkbox or rename an existing one for Text message opt out
we are setting up zoom phone to replace our main number which is a google phone. as part of that, we will be able to text people pretty easily from it, but we are required to have some process that they can opt in or out of text messages.

salesforce has these built in fields
image

We presently have Do Not Contact in use. I'm not sure if we changed the label for one of the above (namely Do Not Call). or it is derived.

We can add another checkbox like Text Opt Out  HasOptedOutofTextMessage
or we can just use the Fax one and rename it. we are never going to use that field.

thoughts? I think we would have to use localization to overload the label.

mguillendev:
# Critical Changes

# Changes

- Added the checkbox Text Messages Approved to Contacts object. 
- Located in the Atlas Contact Layout.
- Added the permission set to view/edit the Text Messages Approved field for Atlas Super Users and Atlas Base

# Issues Closed

#498

### DIFF
--- a/force-app/main/default/permissionsets/AtlasSuperUser.permissionset-meta.xml
+++ b/force-app/main/default/permissionsets/AtlasSuperUser.permissionset-meta.xml
@@ -89,11 +89,6 @@
         <readable>true</readable>
     </fieldPermissions>
     <fieldPermissions>
-        <editable>false</editable>
-        <field>CategoryRule__c.Group__c</field>
-        <readable>true</readable>
-    </fieldPermissions>
-    <fieldPermissions>
         <editable>true</editable>
         <field>CategoryRule__c.Position__c</field>
         <readable>true</readable>
@@ -556,6 +551,11 @@
     <fieldPermissions>
         <editable>true</editable>
         <field>Contact.SystemAccess__c</field>
+        <readable>true</readable>
+    </fieldPermissions>
+    <fieldPermissions>
+        <editable>true</editable>
+        <field>Contact.TextMessagesApproved__c</field>
         <readable>true</readable>
     </fieldPermissions>
     <fieldPermissions>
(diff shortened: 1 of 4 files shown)


force-app/main/default/permissionsets/AtlasSuperUser.permissionset-meta.xml
```diff
@@ -88,11 +88,6 @@
         <field>CategoryRule__c.Group__c</field>
         <readable>true</readable>
     </fieldPermissions>
-    <fieldPermissions>
```
deb761, line 91: This needs to be added back here, also.

deb761 (changes requested):
Looks good except for the missing permission.

force-app/main/default/permissionsets/AtlasSuperUser.permissionset-meta.xml
deb761: My bad.  You're removing a duplicate.

deb761 (approved):

atlasjen:
I see this in the page layout editor
image

But I do not see it in the page layout itself
image

I'm guessing some permission setting ?

atlasjen (changes requested):
see earlier comment on changes needed.

force-app/main/default/permissionsets/AtlasSuperUser.permissionset-meta.xml
mguillendev: Sorry, I don't understand. Do I need to add something?

deb761:
Did you try a hard page refresh?  I see the new field in AtlasSuperUser and the AtlasBase permission sets.

force-app/main/default/permissionsets/AtlasSuperUser.permissionset-meta.xml
deb761: No, when you pulled the code down, it removed an extra field permission.  On first review, I thought you had accidentally removed the permission altogether.

atlasjen:
ok. with a full org redeploy it's there. thanks!

PR #501, opened 2023-10-20
417 Make filtering more visible in file lists

deb761:
# Critical Changes

# Changes
1. Added pill container to show the file filters
2. Added handling of the delete pill to remove the filter
3. Fixed problem in PositionStatusController when date not defined

# Issues Closed


atlasjen:
I love it! so the only question i have is if i pick "blank" as my filter option, it changes to "no filter"
this was a case where i uploaded files w/o a category or type.

image
image

deb761:
That was a case I hadn't considered.  I'll need to think about it.

deb761:
Here's some pics with the solution I came up with.
image
image
image

atlasjen:
That makes sense. I’ll go check it out.

PR #536, opened 2024-07-07
Feature: restrict which records and categories users can see

deb761:
# Critical Changes
Restored logic and fields for sharing rules and limiting categories.
Swapped category rule rule from Upload to ReadOnly so fewer rules need the setting.

# Changes

# Issues Closed

# To test:

1. Deploy to scratch org.  The rest is in the scratch org:
2. Create another user with Salesforce license and Standard User profile.  Use Atlas Minimum if available.
3. Give this new user the Atlas Base & Atlas Volunteer permission sets.
4. Add Relationships to the profile you gave the volunteer user.
5.  Change Account record types in the profile from Master to Organization and Household
6.  In Public Groups. Add All to the other groups.
3. In Public Groups, Add default user to All Group.
4. In Public Groups, Add new user to Volunteer Group.
7. Run method to set sharing fields:
8. In Developer Console > Debug > Open Execute Anonymous Window
1. Enter `PostInstall.setSharingFields();`
1. Press Execute.
1. The log should indicate success.
2. Log In as the volunteer user.
3. This user should be able to see only the Volunteer contacts and their primary and secondary emergency contacts.


atlasjen:
I believe i followed these steps.. other than i am unclear on what level of permissions to add for Relationships. (step #4) 
I logged in as the Volunteer user, but am still able to access all contacts. For example:
image

PR #538, opened 2024-07-13
Defect/error on relationship delete

### LINKED ISSUE #524: Beta12 & staging: permissions on admin (Jen) don't allow deleting relationships (but can create them)
As Jen, i went to delete a related contact, in particular, the Bryan King contact on Cook Megan.
Here is the error I got:

image

deb761:
# Critical Changes

Corrected relationship trigger so that it doesn't crash when deleting relationships.
The problem was it was using a property that is null during delete.  Switched to using the appropriate property, and added tests to verify proper operation.

# Changes

# Issues Closed
#524


atlasjen:
I can't actually see the relationships related list in this scratch org.
Looking at the Atlas Contact Record page, it's not there and i'm not sure how to add it
image

deb761:
You'll need to add permissions for the Relationship object to the System Admin profile.  I'll go ahead and add a permission set for Relationships to the unpackaged part of the project.

atlasjen:
So i went and checked and they are there in the profile.  all of the other expected related objects are shown. 
image
I am also seeing that the page layout is not what i expect for my user.
I double checked though, i am in as user user, which is a sys admin user. and when i looked at the atlas contact layout and previewed what a sys admin profile should see it was correct, but it is not when i am just looking at the contacts itself. ie, demographics and other custom fields are all missing. 
image

deb761:
I think you are getting hit by a change to how cumulus is deploying unpackaged data.  The Page layout settings are not getting deployed.
If you run `sf project deploy start -d force-app -d unpackaged/post` it should deploy the rest.  I'll look into the problem with the auto deploy.

atlasjen:
i tried that.. i still can't see relationships in the related list. 
I can create relationships by adding a quick access link to relationships. i can delete the relationship i created.
I can't do this in prod, so that is a good sign this is working as expected. I'm inclined to say commit it and we can try it in a sandbox

deb761:
Can you approve this PR then?

atlasjen (approved):

atlasjen:
merged and closed

Issue #531, opened 2024-04-19
Add a new field to client logs


atlasjen:
Let's add a checkbox to client status logs

"Did you meet with your TF this week?"

deb761:
Should this also be in the related list?

deb761:
Closing this and picking up additional field in #545

PR #539, opened 2024-07-14
Add field for DidMeetThisWeek to Log and ClientDog to Contact

### LINKED ISSUE #529: Add "Client Dog" as displayed field in Client Related Details
There are so many cases where we want to be able to include the dog's name when emailing a contact and we can't as we have no access to related object fields.

let's add a field "Client Dog" that shows under Client Specific Details.
It would be from Team, much like Client status is and would be the dog's name vs a link to their object.

deb761:
# Critical Changes

# Changes

1. Added DidMeetThisWeek to Client Log
2. Added ClientDog to contact & Client quick action
3. Added setting contact field to TeamTrigger
4. Added new fields to Team permission set
5. Removed group recipe from snowfakery recipe

# Issues Closed
#529 #531


atlasjen:
i tried deploying to a fresh scratch org. I got this error in deploy

image

deb761:
I just pushed a fix.  It should work now.

atlasjen:
the scratch org built successfully, after the update.
I can see the new fields in the object settings for both Log and Contact.

I did need to add the "worked with TF" checkbox to the page layout for logs... but i'm still not seeing it show up when i look at the logs directly, nor is it in the New Client Log form. 
We will also likely want to add it to the log related list under contacts.


However, the atlas settings like client page layout aren't coming through. I don't see the Atlas page layout.
 I tried `sf project deploy start -d force-app -d unpackaged/post`
but while that ran successfully, it didn't make a difference.
ideas?

Client Dog is populated :)

deb761:
I fixed a problem with the profiles being ignored.
As far as "not seeing it show up when you look at logs directly", do you mean when you pull up the record list page?  Are you saying you want it in the default list view?
And do you want it in the related list for Contact?  Would that be for both Client and TF logs?

atlasjen:
I deployed to a new scratch org.
All is good with Client Dog and propagating from Team.


Regarding logs, if i go to edit an existing log, i see the new field:
image


 here is what I see when i go to create a new Client Log: It does not have the field:
image

When i look at the related logs from w/in the contact, it does not show, so i would like it in the default list view:
image

When I go to look at View All, it also doesn't show: 
image

I have no way to edit the fields shown in these list views.
We didn't add this to TF Logs, so it only applies to client logs.

deb761:
So, what I am hearing is:

1. This new field only applies to Client logs
2. You want to see it in the New Client log dialog
3. You want to see it in the Client log related list
4. You want to see it in the default list view (View all)

atlasjen:
yes, that sounds correct!

atlasjen (approved):
looks good!

PR #543, opened 2024-07-21
Feature/file upload no category

### LINKED ISSUE #537: in PR restore restrictions, can't upload files w/o category
This works fine in prod. In this PR:

- upload a file w/o a category or type is successful, but not shown in the related files in the contact.
- can add a file w/ category N/A and no type w/o issue. Saw an error once but can't repro
 
related to #536

deb761:
# Changes
Changed file controller to allow everyone access to files with no category.

# Issues Closed
#537

### DIFF
--- a/force-app/main/default/classes/FileController.cls
+++ b/force-app/main/default/classes/FileController.cls
@@ -236,7 +236,7 @@ global inherited sharing class FileController {
             
             for (Object item : allFiles.items) {
                 FileInfo info = (FileInfo)item;
-                if (allowedCategories.contains(info.cv.Category__c)) {
+                if (allowedCategories.contains(info.cv.Category__c) || info.cv.Category__c == null) {
                     related.items.add(info);
                 }
             }


atlasjen:
this seems to work just fine now.
Curious what the difference of NA vs blank is.. 
with N/A we have a contact form as an option which applies to everyone.
do we want to default to N/A for everything? Approving this PR as it does resolve the issue listed, but not merging it until you comment on what we want to do here.

atlasjen (approved):

deb761:
The difference between N/A and blank, null, empty-string, is just that: it didn't match the search criteria.

Conceptually it can mean they are equivalent, or not, depending on business rules.  Since these are optional fields, we need to rely on Molly and Laura to make sure they set the category properly to prevent others from seeing files.

If we want to be more stringent, we can consider these required fields, or at least Category is, and fall back to "N/A" for those ContentVersions that don't have it set.

atlasjen:
okey dokey. merging it in .

PR #567, opened 2024-09-05
548 add ValidUntil date to Background Checks

### LINKED ISSUE #548: add ValidUntil date to Background Checks
add another field to BackgroundCheck
"ValidUntil"
let's initially populate it with the value in Date + 1 year and put a trigger in place so that when we upload a new background check it populates.

then update the related list for Background Checks that shows on Contacts to also include this column. let's add it to the right of Completed Date

deb761:
# Critical Changes

# Changes
Added a method to BackgroundCheckController to create reminder tasks to start Background Checks for contacts that have `Credit Check Required` or `Background Check Required` and no Background Check within the last 11 months.

Created task Queue called BackgroundCheckAdministrator to assign the Tasks to.
Created BackgroundCheckRenewalYears in Atlas Settings.

# Issues Closed
#548


atlasjen:
how do you recommend i test this one?

deb761:
To test:

1. Set up a few contacts with the System Access of 'Credit check required' and 'Background check required'.
2. Put your user in Background Check Admin queue.
3. Add Background checks to some of the contacts, even to contacts without the system access.
4. Set one background check to over a year ago, one to 11 months ago, and one to 10 months ago.
5. Go to Settings > Apex Classes.
6. Click on Schedule Apex:
 schedule apex

8. Select BackgroundCheckTask
image
9. Setup the task to run at a convenient time.
image
10. Wait for the task to run.
11. Verify you have tasks assigned.

atlasjen:
I've tried this a couple of times but so far the jobs never run. So i haven't been able to test. ideas?

atlasjen:
if it is helpful, i tried bringing in the PR for the CEUs and pushing to the same scratch org and i got this error:

image

deb761:
You can run the reminder method directly. I’ll send you the lines later. Sent from my iPhoneOn Oct 2, 2024, at 10:35 AM, atlasjen ***@***.***> wrote:﻿
if it is helpful, i tried bringing in the PR for the CEUs and pushing to the same scratch org and i got this error:
image.png (view on web)

atlasjen:
reminder to send me info on how to kick this off manually :)

deb761:
`BackgroundCheckController.createReminderTask();`

atlasjen:
i thought you were going to add it to a sandbox since i haven't been able to get these to reliably run in the scratch orgs.

deb761:
I can. I hadn’t because it is creating a task rather than sending an email.

atlasjen:
Ok. let me build that out to a scratch org again and see if I can launch it directly.
Remind me how to launch the code?

atlasjen:
I am testing w/ manually kicking off the task.

1. It makes a new task each time, which means duplicates. I am not sure if that is a result of running from the Execute Anonymous window, but dep on how often the check runs, we wouldn't want to create a ton of duplicate tasks.
2. I successfully have a reminder for someone who 
** has Background Check required
** has an expired (> 12 mo old background check)
3. I do not have a reminder for someone who
** has Background Check & Credit Check required
** has a background check from a bit more than 11 mo ago  (12/8/23) and NO credit check
**I would expect to have a reminder for both**
4. I have an individual that requires a credit check only. 
** I set her credit check to 10 mo ago, but its status was pending. **I didn't not get a task.**
** I changed it to be a bit over a year ago, w/ passed status. This time i did get a task. it says order Background check vs order Credit Check..  that is likely ok as the user can go see what is required and order it.. but it is a bit confusing. **If they needed both, would they get 2 tasks or one?**
5. I tried a case with person w/ Background Check due and the logic seems to be correct
** older than 11 mo = get task
** done more recently than 11 mo; no task.
6. I tried a case w a person w a background check due and the only one there as failed
** I only get a task if the date is older than 11 mo. **I would expect it for any date.**
7. It looks like the due date is in 1 month?  is that based on when the background check expired? or just always 1 month? that should be fine, more just thinking if it is expired, ideally we'd make it due now.. but that gets more complex, so we can just leave it 1 month since once this has been running, in theory they are getting flagged at month 11.

**So.. i think things are looking good when just background check is involved.**
-It looks like some challenges if we add in credit check too.
-Also curious if status comes in the logic at all.

deb761:
1.  I was thinking this would run once a month, hence no checks for existing tasks.
2. I hadn't considered needing both types of background check.
3. The logic is if there hasn't been a background check in the past 11 months.  So if there was one 10 months ago, it will not trigger a reminder.
4. I am not checking status of the Background Check.
5. The current logic would only do one task per contact, and is hardcoded as 'Background check'.  I can, of course, clean that up a bit.

atlasjen:
I think once a month is good and keeping it due w/in a month is good.
We do want to watch for either credit and/or background. their schedules should be the same if both are required, so 1 task is fine. ideally it would say which they need.  most people are background only. a handful are background + credit. a handful are credit only.. well 1 i think.   
i think we keep it simple for now and ignore status. that can be a refinement if needed. it shouldn't stay pending more than a month and if failed, we would presumably have them in  a status where they aren't required.

atlasjen:
ok.. confused. i thought the update you pushed a couple of weeks ago (Sorry!) also added credit check, but looking at the source , i don't think so.

what should happen if:
- they have background or credit check required and there are NO previous bg checks? i would assume it should create a task. i am not seeing that it does.
- i tested a case w/ just a credit check that was done about 15 mo ago and it correctly created a task
- i tested a few cases where it was only a bg check required and check was 12 mo and i am seeing it create a task correctly
- however, in cases where there are both BG and CC's present and required, i am not seeing it create tasks where i expect it to.

atlasjen:
tried with a new scratch org . 
Got the various contacts setup in diff states; added my user to the group.
Then went to trigger the code using an execute anonymous window:
BackgroundCheckController.createReminderTask();

this isn't happy
image

I looked at the present code. I think this needs to be on BackgroundCheckService  not on BackgroundCheckController.

I tried then instead running the following from an execute anonymous window:
BackgroundCheckService.createReminderTask();

image

i looked through the tests, but i am not sure the best way to initiate this.

deb761:
```
BackgroundCheckService service = new BackgroundCheckService();
service.createReminderTask();
```

atlasjen (approved):

PR #575, opened 2024-09-29
546 create jobs to remind clients and facilitators to submit logs

### LINKED ISSUE #546: Datamodel support to Automate emailing log reminders to clients and facilitators when client is in training
We have the use case of wanting to be able to schedule emails for all in-training teams to remind both the client and facilitator to log their status reports.
Clients log weekly. 
Facilitators log every other week.

Maria was looking into how to do this with flows and hit the issue 
- when on the team object, she has no access to the contact who is the facilitator to get their email.
- when on the contact object, the facilitator info is in a related list (relationships) and it's unclear if she can get that out of the flow. She wasn't succeeding at intial attempts.

ideally the emails have some salesforce fields that make it obvious who they apply to.. so some derived fields.
so for facilitator, it would say they need to file their log for client X.  or Team X and the URL would be derived.
for client, it would say they need to file their log for Team X and same deal, the URL would be derived.

ideas? Maria said one way to trigger it was for there to be a boolean like "TeamLogReminder" or "ClientLogReminder"
This would be set based on Team/Client status changing state to or from InTraining.

What is trickier, is we'd want to do the same for the Facilitator .. so if a client is in training, we go to relationships and look for the faciliator relationship and then that is who we active the FacilitatorLogReminder for.. but the checkbox alone here doesn't give us the info on which team it is for.. and some TFs have more than one client.

if there was a way to access the facilitator in the Team, that could work, but since everything pushes down from team to contact, i worry about something populating in reverse. 

Debbie, thoughts?

deb761:
# Critical Changes
Created Repository classes and used System StubProvider to create unit tests for email create and send.
Updated TeamService to use these repositories.

# Changes
Added email templates for the log reminder emails.
Added two Schedulable Apex classes: one to remind clients and the other to remind facilitators.

# Issues Closed
#546


atlasjen:
verified i can add the facilitator to team.
i am using "Choi Gloria" who is a client w/ In Training status and part of a team. I added Cassie Crosby as their facilitator and made sure that Cassie had position of Facilitator and facilitator status of certified-active.

i tried looking in Email Templates to validate those, but don't see any 
image

I also tried seeing if i could find the template via initiating an email, but none show up
image

Just fyi, we have the active email templates we use here in prod:
image
I would say these could go under Contact templates/Renewal templates

what steps should i be taking to test this?

deb761:
Mmm.  The team facilitator needs to be related to the client for the email to be sent.  I set it up to use a template in the package, but I think I should grab it out of Atlas Settings.
The template templates are under `AtlasEmailTemplates`.

To test, 

1. Make sure a Team is `In Training`, and the client has a Team Facilitator relationship with someone.
2. go to Settings > Apex Classes and cilck the `Schedule Apex` button.
3. Select `RemindFacilitatorsToGetCeus`
4. Pick a time to run at least a few minutes in the future.
5. Wait for the job to run.

atlasjen:
ok i found the templates back in classic templates.
are we able to use lightening templates for these? 
yeah, it would be ideal if we grab the name of the template from atlas settings.

I went ahead and scheduled a facilitator one to run at 7pm and changed the email of the facilitator to be one i can access.
I also scheduled a client one to run at 7pm and again made the email a different one i can access.

waiting to see what happens :)

atlasjen:
Not sure why the client one didn't trigger, but here is what i saw for the facilitator.

image

Here is what i set:
image

image

I did not set the relationship on the contact themselves, since my understanding is you are using the one in team.
Back to you.

atlasjen (changes requested):
see comment.

atlasjen:
I wonder if the team object is in scope as this runs.. since the URL it is trying to build is pulling from team.

deb761:
You **told** me to use the relationship, not the team property.  I can modify that of course.

atlasjen:
AH ok cool. i def prefer it w/ the relationship. so is the Facilitator Field under Team one we can remove?
restarting a test w/ the facilitator relationship set

atlasjen:
So it turns out i had the relationship set. however, something is different w/ reciprocal relationships in staging.
If i set it to team facilitator on one side, it should set the other to client. 
However it is setting them to the same value.
so if i make one "team facilitator" that is what it sets the opposite. if i manually edit the opposite, it changes the other.
the reciprocal relationship is set properly.
the only thing i saw different was the field specified for gender to use in reciprocal relationships.
in prod it is "gender" in staging my closest option was gender identity. it shouldn't matter as there is also a neutral case.

I went ahead and made it team facilitator both directions. (I am pretty sure this is what i had before when i tested)
I cleared out the value for "Facilitator" from the Team object.

I scheduled it to run at 10pm, so we shall see what happens when i check in the am.
I am assuming the Client one depends on "client" as the relationship, and this reciprocal issue may be a problem for both.

The client one i scheduled yesterday still has not run.

atlasjen:
note to self- how to run from developer console: 

TeamService service = new TeamService();
service.remindFacilitators();

TeamService service = new TeamService();
service.remindClients();

atlasjen:
I got the emails successfully on 2 diff clients and 2 diff teams using the unicorn sandbox. let's go ahead and commit this.
I would like to update the template to pass another field and i need to test that the team name field we are passing is what i need to go from form assembly to salesforce so let's go ahead and commit knowing we may have an update.

question on this- am i able to use any fields available off of team for the template, or are you manually loading the data for the specific values? i didn't dig in the code to look at which level you have the object available.  if it is the full team object or contact, then ideally i could change the values pulled from it in the email w/o a code change.

let me know. i am going to go ahead and merge this in.

atlasjen (approved):

atlasjen:
actually, i went ahead and approved the changes, but can you do the merge and commit?

Issue #584, opened 2024-11-20
on CEU reminder email, can we include # valid CEUs in that term


atlasjen:
For the email reminder we have, can we include the # of valid CEUs they have that will be applied in that term?
that is one of the most common questions i get.

atlasjen:
update #547. add in a list of CEUs that are applicable in the Active Certification for that contact.

ceu name, program title, date completed, requested CEUs, status

deb761:
I can include the # of CEUs, but the list would require abandoning the template or using some custom templating framework.

atlasjen:
i am open to changing the template to something else. no clue what is involved in a custom template framework. Also open to restructuring the CEU object if we do it soon. i will be needing to import these in the next month, so now is a good time to change it if we want to

Issue #578, opened 2024-10-09
need to add the same Leeway logic to Facilitator Certification Valid until that we did for trainer


atlasjen:
See #571 
this resolved #555  for trainers, however, it does not appear the same logic was added for facilitators.
Can you add it? I think we can use the same renewal leeway for both.

atlasjen:
This isn't working. i just reopened the ticket on certification objects commenting on this.
it works great for trainer but not facilitator,.
Look at "Ms. Adrienne Weeks" in staging
https://atlasassistancedogs--staging.sandbox.lightning.force.com/lightning/r/Contact/003U900000V7hzIIAR/view

deb761:
I'd forgotten to swap out the fields in the action layouts.  Looks correct now.  Check again.

atlasjen:
yes. works now. thanks

Issue #573, opened 2024-09-18
is it possible to add a trigger to set Initial Certification Date based on the 1st Certification Agreement uploaded. this applies to Trainers and Facilitators


atlasjen:
Right now when we upload a document   X Certification Agreement it sets the "Last  X Certification Agreement". where X is facilitator or trainer.

Is it possible to also have it have a trigger to set Initial Certification Date, which I believe already triggers Certification Valid Until?

idea would be if it is the 1st time a document type of Certification Agreement got uploaded for a trainer or facilitator that it would set that Initial date field, but if it would not be updated if another certification agreement was uploaded. we'd only update the "last" date.

atlasjen:
See #586 when you are adding this as it is going to be in the same block of code.

atlasjen:
is it expected behavior that if we change the date on the file that has been uploaded that it does NOT change the last* dates?
if so, will close as i verified the rest.

PR #589, opened 2025-03-26
Feature/586 certification objects

### LINKED ISSUE #586: Certification objects
We need to have a way to associate which CEUs are associated with a given certification window and they might apply to both trainer and ceu.

if we add "Certification" as a custom object in a related list for the contact
* type (trainer, facilitator)
* contact
* status (active, complete)
* start date
* end date
* related CEUs-> link to associated ones in this time period 
* total CEUs (trigger based on if a new CEU is added) <- idea here is we might programmatically add CEUs, so we would want this to update based on that.

end date = last valid until for TF or Trainer.
start date = end date- 1 year

### **creation logic**
- when we set Facilitator Initial Certification Date, we calculate the certification valid until. could we at that time create a new Certification object of type facilitator? setting its start date to initial cert date and end date to valid until.  Same for trainer.
- when we add a new agreement, that updates valid until. .at that point, change status on any previous Certifications to "complete" and create a new Certification with the new dates

- Then, whenever a CEU is added for a contact, it should be associated with any active Certification for that contact (trainer or facilitator).. ie, may be associated with both.

### **post install script**
for every trainer and facilitator who is active, we need to create the current Certifications based on the valid until dates.

### **post deploy**
Jen to remove present CEUs and do bulk import

deb761:
# Changes

1. Created certification object
2. Changed name of ContinuingEducationUnit to Ceu and changed some field names
3. Created CeuCertification connection table to tie CEUs and Certifications
4. Created trigger methods to connect CEUs to active Certifications
5. Created trigger to create Certification when Certification agreement is added
6. Updated reminder message to include current CEU quantity

# Issues Closed
#586, #573, #578, #584


atlasjen:
### Here is what I tested. if in bold, it means we should fix or consider. i added a summary at the bottom and numbers in front of the items that need fixing or discussion.

Most things are working quite well :)

-----------
### TESTED:

- In contact Katilin Braun
-- Added a new certification agreement for a TF and it successfully created the Certification object.
-- added a new ceu for that user to "pending" w/ a date in the certification window
-- changed the status to "approved". the ceu's successfully showed up in the certification object.
-- changed the # of ceu's in the CEU and saved. the certification total ceu's correctly updated.
-- added a new ceu w/ a date before the current certification window. it correctly did not get added to the certification.

**--1.  updated date completed on the new one to be in the window. (left program date out of the window). **CERTIFICATION DID NOT UPDATE****
**-- updated program date on this one to be in the window.**CERTIFICATION DID NOT UPDATE****

**2_UPDATE: I think we should remove one of these dates and just use one.  let's keep "date completed" and leave out "program date"_**

-- added a totally different CEU in the proper time window and Certification did update properly.
-- changed this CEU to be denied and it properly was removed from certification. changed it back to approved and it was properly re-added.

**--3 added position trainer and added a certification agreement w a slightly earlier start date, though with overlap. New certification properly gets added but it does not have any of the CEUs previously entered for its time window. **PREFERRED it would be ideal if it would pull in any CEUs current in an active certification. It is a pretty common use case for someone to start one program w an overlapping window to the other , so this is going to be pretty important. otherwise, we need a way to manually associate ceus w a certification object.****

-- added a new CEU with date completed n one of the certification windows. it properly added to that not the other. It appears this is the date being used vs program date, which is good.
-- added a new CEU w date completed in both certification windows. it was initially pending. it properly didn't add to the certifications. changed it to approved. it properly added to both.
-- reduced the CEU hrs in the above one and it properly subtracted in both certifications
-- increased the CUE hrs in the same one and it properly changed/increased in both certifications

- switched contacts to start fresh. in contact John Meadows
-- added both TF and Trainer as positions
-- added a CEU prior to adding a initial certification agreement to either

**--4 manually edited the Trainer Initial Certification date. A certification agreement was not created. This is also ok as long as when i programatically add a file, it causes the certification to be added.** 

- switched contacts to start fresh. in contact Kevin Lloyd
-- added both TF and Trainer as positions
-- added a CEU prior to adding a initial certification agreement to either
**-- added the initial TF certification agreement w an earlier date, but one the CEU above is in. It was not added to the certification. this is the same issue as item 3 above.**

**-- 5 The date for the initial TF certification agreement i uploaded was old enough that the valid until date is in the past. The certification is created as active.  Should it be?** 

-- added a new TF certification agreement w 3/29 25 as the date. The original date was 2/8/24. The old certification became "complete" and the new "Active" and w/ 2/8/25 as the start date. I think that makes sense. actually, see #7 below. related to this.

**--6 added a CEU that would fit in the window for the now complete certification. it still added to the certification. should it? this could be good as i could see us trying to catch up on documenting ceu's from prior windows. thought i am torn if this should happen as it could lead to us double counting ceus. i think this goes back to we should likely just add the ability to manually associate CEUs w/ a certification.**

**--7 added a CEU with the exact date that was the end date of the previous CEU window and start date of the next. it added to both. it should only go in one or the other.  since we slide the window back for the new certification to be back to the initial certification date, then maybe when we close the previous one we shift the day back 1 day? curious what you think makes sense.**

Image

-- added a trainer certification agreement in the middle of the windows of the TF active and completed.

Image

-- added a ceu w a date in the complete TF and active Trainer. it added it to the completed TF and active trainer certifications. it didn't add to the active facilitator, so that is good.. it didn't double count. i think this is ok.
-- added a TF certification agreement w a date earlier than the current active certification. no certification was added. this sounds correct. it appears the certification valid until also doesn't change on the contact, so i think we are good.

Image

Image

-----------
### SUMMARY TODO:

A. fix items 1,2,3*,7 above  (*see F below, though doing 3 is still ideal)
B. review item 4,5,6 above
C. I can't seem to add Certification to the top nav bar; it is not an available option
D.  when we look at the related CEUs under certification, can you show the same info as we presently show in the CEU detail list
E.  I love that there is  a refresh option next to certifications. can we add that to all related list items? right now it is on certifications, teams and relationships, but not the others, including ceus.
F. Add the ability to manually associate CEUs w/ certifications. It would be add or associate under the related CEUs under a certification. If we do this, then i think we remove the ability for ceus to automatically add to certifications that are closed. this avoids our double counting issues above. and also gives us a way to add ceus that may have been added before the certification was added (so if we don't fix 3, this is a solution)
G. We want to make sure programatically adding CEUs has the same results.

-------------
### THINGS TO DOCUMENT AS BY DESIGN/KNOWN:

A. I tested changing the trainers initial certification date manually (not via the file dialog) and the certification window didn't change. i think this is fine.  
B. I tested changing the trainer's initial certification date by changing the certification agreement file data and the certification window didn't change. again, this is fine.
C. None of this is dependent on trainer or tf status (which is fine).  we would prefilter based on that if needed.

atlasjen (changes requested):
see detailed testing comments.

atlasjen:
in testing, see can associate existing CEU w/ a certification.
- need to filter to only show CEU's associated w the contact.
- in CEU related list on certifications, show more CEU details

atlasjen:
Immediate TODOs
- remove "program date" from CEU
- update code in certification so "completed" certifications don't get updated by trigger, only active
- add in tab for "certification"
- change label on certification form where you add/edit CEU's from "approved" to "hours approved"
- see if can prevent the same CEU object from being added twice to a certification
- add CEU "status" in the CEU detail page under certification. right now it isn't present. (and comment.)  essentially all fields in the normal CEU edit page layout
- On CEU related object: enable sort on Name and Date Completed.
- enable reporting on certification object
- see if can see CEU detail on related list on Certifications

Moving to NEXT 
- adding custom component for CEU related list on Certification
- adding filtering so only CEUs that are associated w/ the contact can get added to a certification
- having certification form that adds new/exising CEUS automatically pull the right number of CEU's vs adding in manually
- if not fixed above, add logic to prevent the same CEU object from being added twice to a certification

deb761:
@atlasjen I have created a test package and put it in the emails sandbox.  https://atlasassistancedogs--emails.sandbox.lightning.force.com/lightning/o/Account/list?filterName=__Recent

Issue #577, opened 2024-10-04
populate Contact.Client Dog based on values in Team


atlasjen:
We added "Client Dog" as a field that shows under client specific details, but there is no automated logic to populate it.
We wanted access to it for emails. 

Can we add logic to team that just like setting client status on the contact it sets the client dog name?

![image](https://github.com/user-attachments/assets/3d20eb41-b77c-4aa5-9c32-acc54540be7f)

deb761:
This is already implemented.  Added a unit test with #594 557

PR #594, opened 2025-04-05
Add X-ray required

### LINKED ISSUE #557: Add "XRay Required" to Team and Dog
On Team, let's add 
"Xray Required" as a checkbox. 
thinking add it to the left of Last Vet XRay report and then move the Last Vet health form down a line.
image

Then, let's display that same value under dog based on what is in team.
Please add the same tooltip that you have on Status in Dog.

image

deb761:
# Critical Changes

# Changes
1. Added X-ray required to team and dog
2. Added setting value in Dog based on team to team trigger

# Issues Closed
#557 #577

### DIFF
--- a/force-app/main/default/classes/teams/TeamService.cls
+++ b/force-app/main/default/classes/teams/TeamService.cls
@@ -90,6 +90,7 @@ public inherited sharing class TeamService {
 
     if (dog != null) {
       dog.Status__c = team.Status__c;
+      dog.XrayRequired__c = team.XrayRequired__c;
       dogRepo.modify(dog);
     }
   }
--- a/force-app/main/default/flexipages/DogRecordPage.flexipage-meta.xml
+++ b/force-app/main/default/flexipages/DogRecordPage.flexipage-meta.xml
@@ -92,6 +92,12 @@
                 <identifier>RecordSpayedNeutered__cField</identifier>
             </fieldInstance>
         </itemInstances>
+        <itemInstances>
+            <fieldInstance>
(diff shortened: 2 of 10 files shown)


atlasjen:
i can see that "xray-required" is in the Team and Dog objects.
I see it is in the Dog Record Page and Team page layouts, but it doesn't show when i look at the dogs or teams 
I also don't see it in the Edit or New form for the team (or dog).
image
image

we can merge it in as maybe it is just a scratch org issue.. but wanted to see what you wanted. .
i do see that the dog name propegates from Team to contact.

deb761:
Can you be more specific about where else you want this field to be visible?

atlasjen:
here is where it is on the dog layout when i go to edit it(bottom left) which is fine.. but it just doesn't show up when i am viewing on a dog. 
Screenshot 2025-04-07 143855
and then  here on the team under Dog Documents
image

atlasjen:
awesome. looks good. can i merge?

atlasjen:
ah, it is still not there for "New" on Team.  it is visible now on Team and Dog.

deb761:
Do you need it on the new Team form?

atlasjen:
it if it is easy to add, yes. if not, we can do it as an edit. in reality, we aren't likely to know at the time we make the new team and will need more info from the client to determine if they will be doing load bearing work on their dog.. so in reality, it's going to be an edit later.

deb761:
Then, let's merge it in.

atlasjen (approved):

PR #595, opened 2025-04-05
Feature/574 initial certification date

### LINKED ISSUE #574: is it possible to add a trigger to set Client Initial certification Date base on the date on the pat classified as iniital?
I was trying to remember if we already do this,but there isn't a text tool comment on the field so i'm not sure.

I'd like Team.Initial Certification Date to be set to the date of the PAT classified as Initial and Passing.
Sometimes one will be uploaded as initial and have a different state, so it should only trigger if it is changed to both initial and passing.

I know we already then have a trigger to set Client.Initial Certification Date from the one in team

deb761:
# Critical Changes

# Changes
- Added logic to set PAT valid until off the most recent PAT, and to set the PAT count off the number of passing PATs.
- Added logic to set InitialCertificationDate off the first ClientCertification agreement.

# Issues Closed
#574


deb761:
That is a valid point, will change it to set on the first certification agreement.

deb761:
Updated

atlasjen:
i tried to pull the latest but it is not showing any changes to pull. is it due to listed conflicts?

deb761:
Maybe.  I'll address them.

atlasjen (approved):

PR #596, opened 2025-04-06
Feature/592 rename lomi to provider form

### LINKED ISSUE #592: We are renaming LOMI to Provider Form
We need to rename a couple of "Categories" if we can.

1) This applies only to clients.
Clients now submit a Medical Provider Form or Mental Health Provider form.
We don't need to track both necessarily, but can we rename LOMI to "Provider Form"

this is not that critical , but it is  an  area ADI reviews and cares deeply about.. we had to  make a lot of changes in this process for them.. so if we can, it would be good.
if we can't rename the internal object, can we rename the display label to "Last Provider Form"

These next two are generic and apply to any position potentially, so should go under "NA" in the category field dependency. 

2) rename "Agreement" to "Participation Agreement" if we can.
Again, if we can't rename the internal object, can we rename the display label to "Last Participation Agreement"
Move to NA in the category mapping.
Add to main detail pane for contact
"Last Participation Agreement" (could do under Last Contact Form)

3) add "Media Release" as an option. 
This would get a new "Last Media Release" date.
and again add to main detail pane for contact.

this is all stuff coming up as we finalize forms for ADI.

deb761:
# Critical Changes

# Changes
1. Changed label for LOMI to Medical Provider Form
2. Added Media Agreement type to N/A category and Last Media Agreement date
3. Added Participation Agreement to N/A category and Last Participation Agreement date

# Issues Closed
#592

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -3,8 +3,8 @@ project:
   name: atlas-dog
   package:
     name: AtlasDogs
-    namespace: dogs
-    api_version: "60.0"
+    namespace: atlas1
+    api_version: "63.0"
     install_class: "PostInstall"
   dependencies:
     - github: "https://github.com/SalesforceFoundation/NPSP"
--- a/force-app/main/default/classes/contacts/PositionStatusController.cls
+++ b/force-app/main/default/classes/contacts/PositionStatusController.cls
@@ -1,5 +1,5 @@
 global with sharing class PositionStatusController {
-  public final static String[] POSITION_LIST = ContactService.POSITION_LIST;
+  global final static String[] POSITION_LIST = ContactService.POSITION_LIST;
 
   @AuraEnabled(Cacheable=true)
   global static List<PositionStatus> getPositionStatusForContact(Id contactId) {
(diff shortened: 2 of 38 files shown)


atlasjen:
1)
Verified that we now have three fields in teams and they set properly with file upload
image

Since you ended up adding that we can track the mental health one and medical one separately, can you please rename 
"Client Provider Form" to "Medical Provider Form"   i think it will be less confusing. (At least the label). ok if the field stays the same.

so "Last Medical Provider Form" vs "Last Client Provider Form" and then in the document type "Medical Provider Form" vs "Provider Form"

2) Verified can now upload Media Form's to contacts. It was missing in the page layout. I added it.
Since we will manually be adding those anyway in prod, this looks good to me :)

atlasjen:
Confirmed label is updated.

we still need to change the DocumentTyoe. should we just do that in prod when this is released?
just making sure that doesn't impact your code setting the last* dates.
Let me know. if i should commit and merge.

deb761:
What do you mean change document type?  The type name is how the code sets the last date.  So I can't change the type name without messing up setting the last date.

atlasjen:
from "Provider Form" to "Medical Provider Form" ? in the ContentVersion object.  I recognize this may be a small code update.

deb761:
Updated label.

atlasjen (dismissed):

PR #618, opened 2025-05-28
Feature/605 initial cert wins

### LINKED ISSUE #605: Initial certification dates/valid until dates change
I just learned from Molly and Laura that they need to be setting Initial certification dates manually.
People can take a long time to get their certification agreements in and Molly wants these to be consistent w when we send them a letter telling them they have certified.  I clearly misunderstood some of this. so.. here are some logic reverts/changes. Sorry! Hopefully this isn't too big of a pain w/ certification.

**For Teams.**
She will manually set "Initial Certification Date".   Let's remove the logic that set it based on the Client Certification Agreement. I think this is just a revert to previous. this would still copy to Client. Initial Certification Date.

**For TF and Trainers**
we will manually set "Initial Certification Date". That should trigger setting
- Valid until = 1 year after Initial Certification Date  (for the 1st time it is set).

Again, i think the above is a revert to previous behavior.
**Then for Certifications**

- Create the 1st certification window based on those dates vs the Certification File Upload date. So the thing to trigger would be setting the initial certification date for the 1st time.

- All other certification logic remains the same as it is now.  (ie, the certification agreement file upload logic that is there now works after the 1st year)

---------
here is the other bit of logic i wanted to check on.
Say someone expires and is now outside of leeway of 30 days.
I believe we
- close any previous certifications
- create a brand new certification w/ a start date of the latest certification agreement and end date 1 year from that.
--  where do we store the new start date? I'd prefer not to stomp on Initial Certification Date. Do we need a new field like Latest Start Date? and then Latest Start Date can always be set to Initial Certification Date automatically so your certification logic above would go by latest start date?

deb761:
# Critical Changes

1. Set certification valid until to 1 year after initial certification date
2. Base setting new valid until to the current one: if a new agreement is received, leeway is based on validUntil.
3. Tested for both Team Facilitator and Trainer contacts.

# Changes

# Issues Closed
#605

### DIFF
--- a/force-app/main/default/classes/certifications/CertificationService.cls
+++ b/force-app/main/default/classes/certifications/CertificationService.cls
@@ -160,22 +160,41 @@ public inherited sharing class CertificationService {
 
   }
 
-  /* This is called when a new certification agreement is received for a contact.
-   * If this is the first certification for the contact (identified by validUntil being null),
-   *   Creates a new certification and exits.
+  /* This is called when a Trainer or Facilitator contact is updated.
+   * If the contact has an initial certification date but not a valid until date:
+   *   Creates a new certification off the initial certification and exit.
+   * If the contact has a certification agreement but no initial certification and no valid until:
+   *   Create a new certification off the agreement and exit.
    * It update the current certification (if there is an active one, otherwise it creates a cert for the just-completed period).
    *   If the new certDate is within the leeway period, it completes the old cert and creates a new one starting when the old ends.
    *   If the new certDate is after the leeway, it completes the old cert and creates a new one starting on the certDate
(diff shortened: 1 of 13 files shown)


atlasjen:
it looks like certification objects are added twice in related objects under Contact.  I assume one should be removed?
Finishing testing the rest. TF's look good :)
image

atlasjen (approved):
all looks good w. the logic for Initial cert dates and certification windows.
Just see my comment re removing the extra certification from Contact. Related in the page layout. good to merge.

deb761:
Sigh, I didn't see it, so added it to the record page.  Which do you like better?  The first or second?

atlasjen:
I think i like the top one better in terms of column order
image

PR #619, opened 2025-05-28
Add check that CEU is not assigned

### LINKED ISSUE #608: handle CEUs that came in after a certification window expired
We accepted this as known issue in 1.10. Let's fix here.

When we create a new active certification for a client /trainer, check to see if there are any CEUs that should be in that new certification window and add them. we need to be smart enough to close the previous one first so if the end/start dates overlap, we aren't duplicate entering.

this will deal w/ the case where a certification lapses and restarts or someone turns in CEUs in the 30 day leeway we give them after their valid until expires.

deb761:
# Critical Changes

# Changes
1. Add check that a CEU is not assigned To another Cert of the same type before adding it.

# Issues Closed
#608

### DIFF
--- a/force-app/main/default/classes/certifications/CertificationRepository.cls
+++ b/force-app/main/default/classes/certifications/CertificationRepository.cls
@@ -1,12 +1,16 @@
 public inherited sharing class CertificationRepository {
   public List<Certification__c> getForContact(Id contactId) {
-    return [
-      SELECT Status__c, Contact__c, End__c, Start__c, Approved__c, Type__c
-      FROM Certification__c
-      WHERE Contact__c = :contactId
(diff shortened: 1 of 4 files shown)


atlasjen (changes requested):
I just deployed to a new scratch org and tested and still see the CEU got added to both certifications:
image

Here is what i did. for someone who was a TF
- set their initial certification date 2/3/2020
- - that correctly set their valid until  2/3/2021
- - that correctly created a certification for 2/3/2020-2/3/2021
- created a ceu dated 2/3/2021
- set the Last Facilitator Cert Agreement date to 2/3/2021
- - that created a new certification, but as you can see above the CEU got added to both.

atlasjen (approved):

PR #625, opened 2025-07-21
Add client to Log layouts

deb761:
# Critical Changes

# Changes
Added Client to the log LWC form.  Changed it so Session type is visible for new client logs.
Added Client to the layouts, although we will also need to do that in production.

# Issues Closed


atlasjen:
that works.

atlasjen (approved):

